PR #2, opened 2020-10-14
Unique rule names for multiple CloudWatch expressions

edudobay:
## Description

A mixture of two cases that were handled individually uncovers an incorrect behavior: when a function has multiple event expressions AND the resulting rule name would be >= 64 characters long, all expressions get the same hash, so only the last expression is scheduled.

In my fix, the index of the expression is used to build the hash, but only if the index is not zero. My rationale for this was preserving hashes for current rules. This seems unneeded (it doesn't seem like it would break anything without caring about this) but may prevent surprises. I'd like some feedback on whether this is an appropriate choice.

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/1727

## Original PR

https://github.com/Miserlou/Zappa/pull/2102

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -2058,15 +2058,29 @@ USE_TZ = True
         )
         self.assertTrue(len(truncated) == 64)
 
-        rule_name = zappa.get_hashed_rule_name(
+        rule_name_index0 = zappa.get_hashed_rule_name(
             event=dict(name="some-event-name"),
             function="this.is.my.dang.function.wassup.yeah.its.long",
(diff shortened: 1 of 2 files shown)


MerreM:
Any chance you can open this on https://github.com/MerreM/Zappa?

edudobay:
Hi @MerreM, sorry it took me so long: https://github.com/MerreM/Zappa/pull/1

MerreM (approved):
Looks okay to me.

edudobay:
Closing as already solved by #1080

PR #945, opened 2021-02-20
add type hint checking

jneves:
## Description
Add type checking with mypy.


monkut:
old pr that has diverged, closing.

PR #947, opened 2021-02-20
added black

jneves:
## Description
Adding black.


mcrowson:
I like this, but it got a little stale, #980 should cover it

PR #953, opened 2021-03-07
Fix two errors on windows

flozi00:
This PR fixes two small errors on windows
First one is that only empty directory's can be deleted and second one is the encoding of readme when installing from master via pip

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -754,11 +754,11 @@ class Zappa:
         archivef.close()
 
         # Trash the temp directory
-        shutil.rmtree(temp_project_path)
-        shutil.rmtree(temp_package_path)
+        shutil.rmtree(temp_project_path, ignore_errors=True)
+        shutil.rmtree(temp_package_path, ignore_errors=True)
         if os.path.isdir(venv) and slim_handler:
             # Remove the temporary handler venv folder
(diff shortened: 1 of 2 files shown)


zappa/core.py
```diff
@@ -754,11 +754,11 @@ def splitpath(path):
...
 
         # Trash the temp directory
-        shutil.rmtree(temp_project_path)
-        shutil.rmtree(temp_package_path)
+        shutil.rmtree(temp_project_path, ignore_errors=True)
+        shutil.rmtree(temp_package_path, ignore_errors=True)
         if os.path.isdir(venv) and slim_handler:
             # Remove the temporary handler venv folder
-            shutil.rmtree(venv)
+            shutil.rmtree(venv, ignore_errors=True)
```
jneves, line 761: Can you share what kind of error these one causes? I wouldn't expect this to be a problem... kind of surprised.

flozi00: sometimes the folder either did not exist or is not empty, thatswhy ignore error to force deletion

monkut:
Since this is old and there have been no other issues raised that seem to be related to this I'm going to close this.
If this is necessary please open an issue with details of the issue and related errrors.

Issue #960, opened 2021-04-07
Retained versions and alias


Yaronn44:
## Context
When a `num_retained_versions` is set in the zappa_settings.py and an alias has been created manually, a boto3 error trigger.

## Expected Behavior
Versions referenced by alias should be skipped from the deletion process.

## Actual Behavior
The following error trigger and the `zappa update` fails :

> botocore.errorfactory.ResourceConflictException: An error occurred (ResourceConflictException) when calling the DeleteFunction operation: Unable to delete version because the following aliases reference it: [test]

## Possible Fix
Add a try/except to avoid the errror to trigger and making the command to fail.

## Steps to Reproduce
1. Set the `num_retained_versions` to something else than `null`
2. Create an alias from the AWS Console on the Lambda
3. Use `zappa update stage`

## Your Environment
* Zappa version used: 0.52.0
* Python version: Python 3.7
* Your `zappa_settings.json`: 
```
"stage" : {
        "app_function": "app.app",
        "project_name": "stage",
        "num_retained_versions": 5
}
```

PR #961, opened 2021-04-07
Add try/except to avoid error on deleting unretained version with alias

Yaronn44:
## Description
If the `num_retained_version` is set in the `zappa_settings.py` and an alias has been created manually from the AWS Console, the zappa update fails if the version the alias is on is an unretained version.

## GitHub Issues
Problem describe in the following issue : #960.

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -1161,7 +1161,12 @@ class Zappa:
             versions_in_lambda.remove('$LATEST')
             # Delete older revisions if their number exceeds the specified limit
             for version in versions_in_lambda[::-1][num_revisions:]:
-                self.lambda_client.delete_function(FunctionName=function_name,Qualifier=version)
+                try:
+                    self.lambda_client.delete_function(FunctionName=function_name,Qualifier=version)
+                except botocore.exceptions.ClientError as e:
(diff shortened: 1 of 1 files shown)


Yaronn44:
Am I wrong or the error is coming from the requirements upgrade done in this PR : #944 ?
It's seems the error is coming from the urllib3 library

techdragon:
@Yaronn44 is the error you're talking about the CI failure? 

Also, this looks like a pretty sensible addition to me, but I don't have power to merge this.

Yaronn44:
@techdragon yes it's the one !
I know, but Zappa doesn't manage alias at all and this is a really blocking behavior when you have alias managed alongside Zappa... But I think #567 could be a cool feature. I'll definitely need something like this soon so I'll maybe take a look and see what I could do to improve Zappa on this.
Apart from that not having the `num_retained_version` setted mean you can have the AWS Lambda limit of 75 Gb exceeded and that's not really cool either.

Maybe adding a warning or something could be a bit better ?

hellno:
Hi I'm closing this PR as it's 747 days old. 

This is just a housekeeping effort so we can focus on recently opened PRs.
If you want to pick this back and get it merged: 
please reply here or ping me in Zappa Slack and we can get it done! 🔥

PR #962, opened 2021-04-07
Async calls with specific Lambda alias/version

Yaronn44:
## Description
It's now possible to give a specific Lambda alias or version to call with async tasks.
If nothing is passed the lambda called will be on the same version than the calling lambda instead of $LATEST everytime.

## GitHub Issues
Link to this issue #785


monkut:
@Yaronn44 
Can you run `make black` / `make isort` on this?

PR #967, opened 2021-04-18
Update/deploy a lambda using a Docker container image

ian-whitestone:
# Migrated from https://github.com/Miserlou/Zappa/pull/2192

## Background

As discussed in #2188, AWS recently [announced container image support](https://aws.amazon.com/blogs/aws/new-for-aws-lambda-container-image-support/) for AWS Lambda. This means you can now package and deploy lambda functions as container images, instead of using zip files. The container image based approach will solve a lot of headaches caused by the zip file approach, particularly with file sizes (container images can be up to 10GB) and the dependency issues we all know & love.

## What does this PR do?

1. Make the `zappa update` and `zappa deploy` commands accept a new `docker_image_uri` parameter. `docker_image_uri` must point to an image in Elastic Container Registry (ECR) that complies with [these guidelines](https://docs.aws.amazon.com/lambda/latest/dg/images-create.html). Instructions on how to configure this docker image will be provided in an accompanying blog post.
2. Add a new `save-python-settings-file` CLI command, which will be used to generate & save the `zappa_settings.py` file that must be included in your docker image

## Reproducible example
See this [repo](https://github.com/ian-whitestone/zappa-serverless-docker) which contains an example `Dockerfile` that can be used for testing.

## Discussion points for reviewers

* Test cases I added
* [Accompanying blog post](https://ianwhitestone.work/zappa-serverless-docker/) we can point users to for a more in-depth explanation/overview
* Any other key zappa workflows that should be updated and/or tested in conjunction with this? I've tested the `zappa update, `zappa deploy`, `zappa rollback` commands for both creating from a zip file (traditional method) and from a docker image (new method)
* If you would like this PR to be a bit smaller, I can pull out some of the stuff (new `save-python-settings-file` command, wait until lambda is ready function, etc.)  into separate PRs

## Future Work & Next Steps
* Add functionality to zappa to create new docker images. We've lost some of the zappa magic since you can no longer do a "1-line create & deploy" with `zappa deploy`, you need to first create your own docker and then deploy. 
* Implement `zappa rollback`, see [this comment](https://github.com/Miserlou/Zappa/pull/2192/files#r550607530) for details

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -1059,9 +1059,7 @@ class Zappa:
             # it's actually "US Standard", or something.
             # More here: https://github.com/boto/boto3/issues/125
             if self.aws_region == "us-east-1":
-                self.s3_client.create_bucket(
-                    Bucket=bucket_name,
-                )
+                self.s3_client.create_bucket(Bucket=bucket_name,)
             else:
                 self.s3_client.create_bucket(
                     Bucket=bucket_name,
@@ -1243,8 +1241,7 @@ class Zappa:
 
         if concurrency is not None:
             self.lambda_client.put_function_concurrency(
-                FunctionName=resource_arn,
-                ReservedConcurrentExecutions=concurrency,
+                FunctionName=resource_arn, ReservedConcurrentExecutions=concurrency,
             )
 
         return resource_arn
@@ -1284,8 +1281,7 @@ class Zappa:
         #          https://github.com/Miserlou/Zappa/issues/1823
         try:
             response = self.lambda_client.get_alias(
-                FunctionName=function_name,
-                Name=ALB_LAMBDA_ALIAS,
+                FunctionName=function_name, Name=ALB_LAMBDA_ALIAS,
             )
             alias_exists = True
         except botocore.exceptions.ClientError as e:  # pragma: no cover
@@ -1302,8 +1298,7 @@ class Zappa:
 
         if concurrency is not None:
             self.lambda_client.put_function_concurrency(
-                FunctionName=function_name,
-                ReservedConcurrentExecutions=concurrency,
+                FunctionName=function_name, ReservedConcurrentExecutions=concurrency,
             )
         else:
             self.lambda_client.delete_function_concurrency(FunctionName=function_name)
@@ -1495,9 +1490,7 @@ class Zappa:
         """
         print("Deleting Lambda function..")
 
-        return self.lambda_client.delete_function(
-            FunctionName=function_name,
-        )
+        return self.lambda_client.delete_function(FunctionName=function_name,)
 
     ##
     # Application load balancer
@@ -1619,12 +1612,7 @@ class Zappa:
         kwargs = dict(
             # TODO: Listeners support custom ssl certificates (Certificates). For now we leave this default.
             Certificates=[{"CertificateArn": alb_vpc_config["CertificateArn"]}],
-            DefaultActions=[
-                {
-                    "Type": "forward",
-                    "TargetGroupArn": target_group_arn,
-                }
-            ],
+            DefaultActions=[{"Type": "forward", "TargetGroupArn": target_group_arn,}],
             LoadBalancerArn=load_balancer_arn,
             Protocol="HTTPS",
             # TODO: Add option for custom ports
@@ -2121,12 +2109,7 @@ class Zappa:
         """
         self.apigateway_client.update_rest_api(
             restApiId=api_id,
-            patchOperations=[
-                {
-                    "op": "replace",
-                    "path": "/minimumCompressionSize",
-                }
-            ],
+            patchOperations=[{"op": "replace", "path": "/minimumCompressionSize",}],
         )
 
     def get_api_keys(self, api_id, stage_name):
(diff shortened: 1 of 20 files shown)


zappa/core.py
```diff
@@ -1198,29 +1202,34 @@ def update_lambda_configuration(    self,
...
-            Layers=layers
-        )
+        }
+
+        if not is_docker_deployment:
+            kwargs['Handler'] = handler
+            kwargs['Runtime'] = runtime
+            kwargs['Layers'] = layers
+
+        response = self.lambda_client.update_function_configuration(**kwargs)
```
techdragon, line 1232: Any reason for not simplifying this down to

```python
        lambda_aws_config = self.lambda_client.get_function_configuration(FunctionName=function_name)
        if "Environment" in lambda_aws_config:
            lambda_aws_environment_variables = lambda_aws_config["Environment"].get("Variables", {})
            # Append keys that are remote but not in settings file
            for key, value in lambda_aws_environment_variables.items():
                if key not in aws_environment_variables:
                    aws_environment_variables[key] = value

        kwargs = {
            'FunctionName': function_name,
            'Role': self.credentials_arn,
            'Description': description,
            'Timeout': timeout,
            'MemorySize': memory_size,
            'VpcConfig':vpc_config,
            'Environment':{'Variables': aws_environment_variables},
            'KMSKeyArn': aws_kms_key_arn,
            'TracingConfig': {
                'Mode': 'Active' if self.xray_tracing else 'PassThrough'
            },
        }

        if lambda_aws_config['PackageType'] != 'Image':
            kwargs.update({
                'Handler': handler,
                'Runtime': runtime,
                'Layers': layers,
            }

        response = self.lambda_client.update_function_configuration(**kwargs)
```

I could probably get fancy with an inline ternary dict unpacking in the original function call but I like the readability of the `kwargs` definition here, and chose to match that by passing `kwargs.update()` a dictionary so the keys all look the same.

techdragon (changes requested):
While I cant merge this, it does look quite good to me. I did just have the one small stylistic thought while reading through.

zappa/core.py
ian-whitestone: No particular reason, I agree this is more readable. I will update

Quidge (commented):
I was wondering when Zappa would support those new AWS Docker features!

Some non-functional changes can be removed -- I think your editor may have accidentally modified some docstrings (unless you meant to do that).

techdragon:
> I was wondering when Zappa would support those new AWS Docker features!
> 
> Some non-functional changes can be removed -- I think your editor may have accidentally modified some docstrings (unless you meant to do that).

Nice catch on the docstring changes @Quidge. The blank lines in the docstrings should be there. PEP 257 docstring formatting specifies using blank lines to break up multi-line docstrings the way they are in the code currently. https://www.python.org/dev/peps/pep-0257/#id17 

Can you touch this up as well @ian-whitestone?

ian-whitestone:
>Can you touch this up as well @ian-whitestone?

@techdragon @Quidge I'm just waiting on a final review from @jneves. I will fix those up together with any other changes requested by @jneves

onimitch:
As someone about to get started with Zapper I just want to say a big thank you for doing this. Support for Docker was a bit of a sticking point so I'm really thankful you've spent the time to add it and PR it @ian-whitestone

ian-whitestone:
@jneves do you think we'll be able to move this forwards?

onimitch:
By the way, in the example on your blog, in the Dockerfile you copy in the zappa handler, but I think you can avoid this by referencing the handler like this?

```Dockerfile
CMD [ "zappa.handler.lambda_handler" ]
```

onimitch:
One gotcha (sort of) I found with the docker approach is if you use `environment_variables` in the Zappa file, they won't update unless you rebuild the docker image. To get around this, I used `aws_environment_variables` which means they don't need to be included in the docker image. I don't think there's really anything that can be done about this but might be worth mentioning in the docs when talking about environment variables.

ian-whitestone:
Thanks @onimitch , i'll test out both of those things / update the post when I come back to this.

Right now @jneves is the only maintainer of this project and I haven't heard from him in a few months, so TBH i'm not sure if this will get merged unfortunately.

onimitch:
> ```dockerfile
> CMD [ "zappa.handler.lambda_handler" ]
> ```

Update on this.
It breaks the warm up callback because that's set to look for `handler.keep_warm_callback`

So it seems like for now, copying in the handler is the best route.

anush0247:
@ian-whitestone good work, we are looking forward for this feater

ian-whitestone (commented):
FYI @mcrowson I ran `black zappa tests` and it reformatted a bunch of test files...so this PR will include some changes un-related to the docker functionality

zappa/core.py
ian-whitestone: Updated.

mcrowson:
Had a conversation on Slack with Ian, putting summary here for posterity. 

First we're working through an issue with Travis CI where they are not running our tests, so hopefully that will be addressed soon and we can see all of the tests passing.

The second is I have a hard time mixing arguments via the CLI and arguments via our config file. I made #988 to capture an enhancement to allow for any of them. I asked Ian to implement this feature configuration through the settings file instead of a CLI argument. CI/CD systems could use something like sed to update the config file if the tag changes and is not statically set.

mcrowson:
closed/reopened to try and trigger Travis. Nothing to  see here folks.

onimitch:
FWIW I've been using Ian's fork for a good few weeks now and not ran into any issues so far deploying using Docker. Anecdotal I know, but just thought I'd put it out there :D Thanks again for adding this to Zappa @ian-whitestone

ArtikUA:
It works like a charm even with public.ecr.aws/lambda/python:3.8 !

piotrdrazikowski:
In case someone wants to migrate their zip-based project to a docker image, remember to remove `"slim_handler": true` from the `zappa_settings.json` file, otherwise zappa will still download old code from s3 bucket. Just learned it the hard way.

Issue #908, opened 2021-02-20
[Migrated] Update BINARY_SUPPORT to use Content-Encoding to identify if data is binary


jneves:
Originally from: https://github.com/Miserlou/Zappa/issues/2170 by [monkut](https://github.com/monkut)

## Description

Alternative solution to https://github.com/Miserlou/Zappa/issues/2080 intended to allow BINARY and text by making use of Content-Encoding header in the response.

When using _whitenoise_ for caching, which provides compression, binary types may include mimetypes, "text/", "application/json":

- response.mimetype.startswith("text/")
- response.mimetype == "application/json"

Assuming that Content-Encoding will be set (as whitenoise apparently does) this allows compression to be applied to assets which have mimetypes "text/" and "application/json", while allowing for uncompressed "text/" and "application/json" still to be served.

About Content-Encoding:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

Not sure if this works for all use cases, but it currently appears to work for my workcase where _whitenoise_ is performing compression of text/json files intended for caching, while still allowing text for html.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/2080

Quidge:
Our team also ran into this issue. We forked the Zappa repo and merged this older PR: [`bfcb649` (#2141)](https://github.com/Miserlou/Zappa/pull/2141/commits/bfcb6490dc56ba9533622861ab999d9cabe67f06).

@jneves Is there anything you would like to see that could help this older PR code to get over the line? AFAICT it didn't pass original CI because it decreased coverage by `0.03%`.

I'd be happy to add some tests for this scenario and increase coverage.

cdruck:
This also causes failure when it is useful to compress JSON API responses ...

monkut:
released in 0.56.0, closing.

PR #971, opened 2021-04-28
Update BINARY_SUPPORT to use Content-Encoding to identify if data is binary

Quidge:
## Description
This PR re-implements the same functional change that was shown in a [PR from the pre-fork Miserlou/zappa repository.](https://github.com/Miserlou/Zappa/pull/2170)

The change in that PR didn't make it into the Zappa 52 release from what I can tell. My company has been using that PR's code (in combination with the `flask_compress` library) for a couple weeks without issue so I'm opening this PR that copies that code.

Additional tests are added for coverage. I'm not confident that the test organization that I used is correct or idiomatic for this codebase -- please critique and I'll change if wanted!

## GitHub Issues
#908 

### Checklist I found in the contributors documentation:

Before you submit this PR, please make sure that you meet these criteria:

* Did you read the [contributing guide](https://github.com/zappa/Zappa/#contributing)?
	* Yes.

* If this is a non-trivial commit, did you **open a ticket** for discussion?
	* No, a pre-existing ticket [can be found here.](https://github.com/zappa/Zappa/issues/908)

* Did you **put the URL for that ticket in a comment** in the code?
	* Yes.

* If you made a new function, did you **write a good docstring** for it?
	* NA.

* Did you avoid putting "_" in front of your new function for no reason?
	* NA.

* Did you write a test for your new code?
	* Yes.

* Did the Travis build pass?
	* Yes.

* Did you improve (or at least not significantly reduce)  the amount of code test coverage?
	* Yes.

* Did you **make sure this code actually works on Lambda**, as well as locally?
	* Yes. My company has been using these functional changes in a fairly standard lambda environment for a couple weeks.

* Did you test this code with all of **Python 3.6**, **Python 3.7** and **Python 3.8** ? 
	* I did not test personally, but the CI job that ran on this PR [completed successfully after running the suite in 36, 37, and 38.](https://travis-ci.com/github/zappa/Zappa/builds/224510259)

* Does this commit ONLY relate to the issue at hand and have your linter shit all over the code?
	* The changes introduced in the commits are not outside the scope of the issue.

### DIFF
--- a/tests/test_handler.py
+++ b/tests/test_handler.py
@@ -226,10 +226,68 @@ class TestZappa(unittest.TestCase):
         self.assertEqual(response["statusCode"], 500)
         mocked_exception_handler.assert_called()
 
-    def test_wsgi_script_binary_support_base64_behavior(self):
+    def test_wsgi_script_binary_support_with_content_encoding(self):
         """
-        With Binary Support enabled, response mimetypes that are not text/* or application/json will be base64 encoded
+        Ensure that response body is base64 encoded when BINARY_SUPPORT is enabled and Content-Encoding header is present.
+        """  # don't linebreak so that whole line is shown during nosetest readout
+        lh = LambdaHandler("tests.test_binary_support_settings")
+
+        text_plain_event = {
+            "body": "",
+            "resource": "/{proxy+}",
+            "requestContext": {},
+            "queryStringParameters": {},
+            "headers": {
+                "Host": "1234567890.execute-api.us-east-1.amazonaws.com",
+            },
+            "pathParameters": {"proxy": "return/request/url"},
+            "httpMethod": "GET",
+            "stageVariables": {},
+            "path": "/content_encoding_header_json1",
+        }
+
+        # A likely scenario is that the application would be gzip compressing some json response. That's checked first.
+        response = lh.handler(text_plain_event, None)
+
+        self.assertEqual(response["statusCode"], 200)
+        self.assertIn("isBase64Encoded", response)
+        self.assertTrue(is_base64(response["body"]))
+
+        # We also verify that some unknown mimetype with a Content-Encoding also encodes to b64. This route serves
+        # bytes in the response.
+
+        text_arbitrary_event = {
+            **text_plain_event,
+            **{"path": "/content_encoding_header_textarbitrary1"},
+        }
+
+        response = lh.handler(text_arbitrary_event, None)
+
+        self.assertEqual(response["statusCode"], 200)
+        self.assertIn("isBase64Encoded", response)
+        self.assertTrue(is_base64(response["body"]))
+
+        # This route is similar to the above, but it serves its response as text and not bytes. That the response
+        # isn't bytes shouldn't matter because it still has a Content-Encoding header.
+
+        application_json_event = {
+            **text_plain_event,
+            **{"path": "/content_encoding_header_textarbitrary2"},
+        }
+
+        response = lh.handler(application_json_event, None)
+
+        self.assertEqual(response["statusCode"], 200)
+        self.assertIn("isBase64Encoded", response)
+        self.assertTrue(is_base64(response["body"]))
+
+    def test_wsgi_script_binary_support_without_content_encoding_edgecases(
+        self,
+    ):
         """
+        Ensure zappa response bodies are NOT base64 encoded when BINARY_SUPPORT is enabled and the mimetype is "application/json" or starts with "text/".
+        """  # don't linebreak so that whole line is shown during nosetest readout
+
         lh = LambdaHandler("tests.test_binary_support_settings")
 
         text_plain_event = {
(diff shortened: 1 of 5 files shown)


paulnicolet:
Awesome, thanks!

Is there anything preventing to merge this ?

Quidge:
@paulnicolet There's nothing blocking that I'm aware of. @jneves could this be merged?

colinhoernig:
Hi @jneves -- we're using this code from @Quidge in a forked version of Zappa because of a need to enable compression, and it's working very well for us.  If this PR could be merged in, that would allow us to move off of our fork and back onto the official Zappa project which we're really looking forward to so that we don't have to maintain our own fork. 🙂

Quidge:
@jneves rebased this PR atop the (great!) newer black changes to the codebase.

travnels:
@jneves is this getting any movement? Anything I can do to help get this merged?

travnels:
@jneves we've been using this code on a fork (https://github.com/tackle-io/Zappa) for over 3 months and are eager to get it merged. This enhancement has helped with large payload that exceed the lambda max response size (6 MB), since it allows for us to zip the response in the lambda. Let me know if there's anything I can do to help getting the approved and merge. Thanks!

colinhoernig:
@jneves It's been several months -- we'd really love to get this merged in so we can cut off of our own Zappa fork.

tests/utils.py
```diff
@@ -73,3 +74,14 @@ def stub_open(*args, **kwargs):
 
     with patch("__builtin__.open", stub_open):
         yield mock_open, mock_file
+
+
+def is_base64(test_string: str) -> bool:
+    # Taken from https://stackoverflow.com/a/45928164/3200002
+    try:
+        return (
+            base64.b64encode(base64.b64decode(test_string)).decode("utf-8")
```
javulticat, line 83: Nit: `"utf-8"` is redundant here in Py3 and can be safely removed

zappa/handler.py
```diff
@@ -583,7 +583,17 @@ def handler(self, event, context):
...
-                        if (
+                        if settings.BINARY_SUPPORT and response.headers.get(
+                            "Content-Encoding"
+                        ):
+                            # We could have a text response that's gzip
+                            # encoded. Therefore, we base-64 encode it.
+                            # Related: https://github.com/zappa/Zappa/issues/908
+                            zappa_returndict["body"] = base64.b64encode(
+                                response.data
+                            ).decode("utf-8")
```
javulticat, line 594: Nit: `"utf-8"` is redundant here in Py3 and can be safely removed

zappa/handler.py
```diff
@@ -583,7 +583,17 @@ def handler(self, event, context):
                         )
 
                     if response.data:
-                        if (
+                        if settings.BINARY_SUPPORT and response.headers.get(
+                            "Content-Encoding"
```
javulticat, line 587: The test case docstring says:

> With Binary Support enabled, response mimetypes that are not text/* or application/json will be base64 encoded

But wouldn't this base64 encode all responses so long as they have a `Content-Encoding` header (even if its mimetype is `text/*` or `application/json`)?  Perhaps I'm misunderstanding, but the logic here doesn't seem to match that docstring.

tests/test_handler.py
```diff
@@ -223,6 +225,90 @@ def test_exception_handler_on_web_request(self):
         self.assertEqual(response["statusCode"], 500)
         mocked_exception_handler.assert_called()
 
+    def test_wsgi_script_binary_support_base64_behavior(self):
```
javulticat, line 228: Could this be broken into a unit test per relevant scenario to make it more clear what is the expected behavior in each scenario?  Eg. one for `application/json`, one for `text/*`, one for `foo/bar`, etc.

Quidge: I think that's a great idea. I was trying to match the style of the tests when I wrote them, but I do think that style feels claustrophobic. Will do.

zappa/handler.py
Quidge: Your understanding was correct. I was misreading the original PR. As written, the code will encode to b64 if BINARY_SUPPORT is enabled and a `Content-Encoding` header is present. If `Content-Encoding` isn't present, then it may still b64 encode under some more specific conditions.

While breaking up the tests to give categories of behavior it's own function, I changed some of the test documentation language to reflect this as well (https://github.com/zappa/Zappa/pull/971/commits/f096f4c16f367e6c439ce0bc47a9f58287a05eaa#diff-3f86b181ab338949acfc1d27ff975b666acd3f0b016eb4bcce7c79fab09e463bR230-R233).

tests/utils.py
Quidge: Removed the switched `.decode("utf-8")` -> `.decode()` in a rebase. Thanks!

zappa/handler.py
Quidge: Removed the switched `.decode("utf-8")` -> `.decode()` in a rebase. Thanks!

zappa/handler.py
Quidge: Your understanding is correct. The docstring text for the test is misleading. Good catch!

I've updated the tests and the language in [`eb2b50f` (#971)](https://github.com/zappa/Zappa/pull/971/commits/eb2b50f65acb1b04d2a3f3d74c5de6ae919c7fe1).

tests/test_handler.py
Quidge: Tests have been divided up into three test methods based on what they're checking for:
- `test_wsgi_script_binary_support_with_content_encoding`
- `test_wsgi_script_binary_support_without_content_encoding_edgecases` (where the mimetype is inspected)
- `test_wsgi_script_binary_support_without_content_encoding`

Quidge:
@javulticat would you please re-review when you get a chance?

Behavior I started thinking about when fiddling with this: what do you think should happen when a csv file is sent back to the client, say via `flask.send_file()`?

`flask.send_file()` will not add a `Content-Encoding` header to the response, but the mimetype _would_ start with `text/` if the filename ends in `.csv` or if the mimetype is passed manually to the function.

The end result is that a file uploaded through `flask.send_file('some/file/path.csv')` will not be base64 encoded due to the conditional logic that's being added in this PR.

I haven't seen any issues raised around this in the last year or so, but it seems wrong. The csv file would be binary so I think we'd want to encode as base64 for that too. Thoughts?

Quidge:
:| I'm trying to 'resolve' a pending unresolved conversation because it appears to be a merge blocker:
Screen Shot 2021-12-07 at 12 09 14 PM

I addressed this with an update to the test files then rebased my changes to keep the PR to two commits (that may have been a mistake -- I was trying to keep the commits atomic).

Now I can no longer reference the conversation to mark it as resolved.
Screen Shot 2021-12-07 at 12 09 09 PM

Not sure what to do here?

javulticat:
@Quidge conversation resolved - thanks for improving the tests!

I'm not sure I fully understand the question being asked around CSVs.  Could you try restating?

One point that may or may not help answer the question is that a CSV is a plain text file.  So I'd imagine the handling of CSVs (`text/csv`) should probably be the same as the handling of any other plain text file (`text/plain`).

zeevt:
CSV (and plain text) can be in any character encoding, it is not guaranteed that it's ASCII or UTF-8. 
If just `.decode()` works, great.
If there is a header like `Content-Type: text/csv; charset=windows-1252` (there usually isn't) then it can be decoded to `str` without relying on default charset.
But if neither of those is true, then you have `bytes` that should be base64 encoded.

javulticat:
@zeevt that's my understanding as well, so I believe it would be fair to say that this is a potential concern for any text file, correct?  I was confused because @Quidge's message implied this was a concern specific to CSVs (which I definitely may have just misinterpreted - sorry!)

If so, would a solution be to sniff the encoding of plain text and use that to decode it using the proper encoding (and, if that fails, fall back on using base64-encoded bytes)?  Something like:

```python
import chardet

# If plain text
try:
    encoding = chardet.detect(response.data)["encoding"]
    decoded = response.data.decode(encoding)
except Exception:
    # base64 encode the bytes
```

?

zeevt:
@javulticat You understood me correctly, yes. I have two concerns with this, now that I thought about it a tiny bit more:

1. Maybe someone is trying to return an http response with "content-type: text/foo" and with a body encoded with whatever encoding on purpose, and the proposed code would decode it and re-encode it in UTF-8 if chardet guesses correctly. This would be a bug. They should be able to ensure the exact bytes they intended are received by the HTTP client, without forcing application/octet-stream content type.

2. How fast is `chardet`? I would hesitate running it on every response without measuring first.

I now think it's only safe to decode `bytes` to `str` if it actually decodes as UTF-8.

monkut:
Migrated to:
https://github.com/zappa/Zappa/pull/1155

closing

Issue #810, opened 2021-02-20
[Migrated] Slow Cold Starts Becasue App Code Loaded Inside of Lambda Handler


jneves:
Originally from: https://github.com/Miserlou/Zappa/issues/1974 by [ElMoselYEE](https://github.com/ElMoselYEE)

## Context
For my Flask + Zappa setup, I noticed that the app code gets loaded inside of the lambda_handler.  This prevents us from taking advantage of the bumped up horsepower that Amazon providers for free during the initialization phase. 

`self.app_module = importlib.import_module(self.settings.APP_MODULE)`

By hand-modding my code to load my app outside of Zappa's lambda_handler middleware, I was able to reduce my cold start time from 7 seconds down to 1 second.

## Expected Behavior
Take advantage of performance boost for code outside of Lambda handler by importing app code before lambda_handler is reached.

## Actual Behavior
App code imported within Lambda handler, which wastes valuable time.

## Possible Fix
I don't have one 

## Steps to Reproduce
1. My Flask app is available at app.py
2. Use zappa package to create a .zip file, deploy it manually
3. Run Function: takes 7 seconds.
4. Open up .zip file and add "import app" to line 2 of handler.py
5. Deploy function manually
6. Run function: takes 1 second.

I can repeat this process and the timing is very repeatable.

## Your Environment
* Zappa version used: 0.48.2
* Operating System and Python version: Python 3.6
* The output of `pip freeze`: 
my requirements.txt is:
`flask
PyMySQL
zappa
aws-xray-sdk==2.4.2`

* Link to your project (optional):
* Your `zappa_settings.py`: 
`{
    "live": {
        "app_function": "app.flask_app",
        "aws_region": "us-east-1", 
        "profile_name": "default", 
        "project_name": "flask-app", 
        "runtime": "python3.6",
        "s3_bucket": "zappa-380dxekze",
        "context_header_mappings": {
            "x-authorized-dealers": "authorizer.dealers",
            "x-authorized-groups": "authorizer.groups"
        }
    }
}`

monkut:
It appears that this was handled by the pr below:
https://github.com/zappa/Zappa/pull/982

closing.

PR #982, opened 2021-05-24
Coldstart can be drastically reduced by calling LambdaHandler externally

xncbf:
## Description
 
Coldstart can be drastically reduced by calling LambdaHandler externally.

### Before
image
image

### After
image
image
image

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/1974
https://github.com/zappa/Zappa/issues/810


xncbf:
Test code is failing because `settings_name` does not exist. Is there any way to operate without `settings_name`?

Yaronn44:
I'm really interesting in this feature.
How many invocations did you do in the `Before` part ?

xncbf:
@Yaronn44 I'm not sure because I didn't capture it, but it's definitely less than 0.1k

xncbf:
The invocation number is meaningless btw

Yaronn44:
How does it work under the hood ? Why does calling the class like that improve the perfs ?

xncbf:
I don't know the detailed principle, but there seems to be some delay in the first invocation after the cold start.

xncbf:
I need help to pass the test.

flanaman:
@Yaronn44  The "execution context" is kept around for subsequent lambda invocations:
https://medium.com/api-api-joy-joy/unleashed-lambdas-outside-the-handler-4e5fc27cb7b3

See "Take advantage of execution environment reuse to improve the performance of your function":
https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html

I've had an issue with zappa opening too many simultaneous database connections so I'm hoping that this feature will solve that problem

flanaman:
> @Yaronn44 The "execution context" is kept around for subsequent lambda invocations:
> https://medium.com/api-api-joy-joy/unleashed-lambdas-outside-the-handler-4e5fc27cb7b3
> 
> See "Take advantage of execution environment reuse to improve the performance of your function":
> https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html
> 
> I've had an issue with zappa opening too many simultaneous database connections so I'm hoping that this feature will solve that problem

Oh, I didn't see that singleton stuff at the top of the `Handler` class. I guess there might be a slight time gap between module import time and the handler invocation?

MunsuPark:
@xncbf  If you check x-ray in api gateway, you can see this commit is not working. If you are using `slim_handler` option, you could found that `load_remote_project_archive` takes 7-10 seconds. This function called when AWS lambda's concurrency is increased.

xncbf:
@MunsuPark I've tried several things.
I tried reducing the size of the package to less than 50MB and 'slim_handler=false'. but still got the same symptoms.
I'm still using `slim_handler` and it reduces coldstarts regardless of this.

xncbf:
In addition, I forked this branch and applied it to the production environment and have been using it for a week. There are a few other minor issues, but the cold start has definitely been reduced.

cometurrata:
I'd really love to see that feature out  :slightly_smiling_face: 

> I need help to pass the test.

You could do something like: 
```
if os.environ.get("INSTANTIATE_LAMBDA_HANDLER_ON_IMPORT"):
    LambdaHandler()
```

This would become something optional through the ENV variables :shrug:   WDYT ?

xncbf:
I received a personal email from @MunsuPark on June 3rd stating that my commit was wrong.
The coldstart I measured was monitoring only the lambda, but when I looked at the overall flow (api-gateway > lambda), it still looked like the coldstart remained.
Clearly, the cold start still seems to exist. However, in the existing method, lambda initailize seems to be executed when the user calls the api.
When LambdaHandler() is called externally, the keep_warm function appears to take over.
This is pretty useful.
By calling the keep_warm function periodically, the system can experience the cold start that the user experiences instead.

This is my 6 hour x-ray
-api-gateway (Most users experience latency within 1.2 seconds)
image

-lambda-function (still have coldstart.)
image

This is the log of api-gateway before commit is applied.
There were users who experienced intermittent but 4-8 second latencies.
image

This is the api-gateway log up to now after commit is applied.
image

cspollar:
I've been running a custom fork with a`LambdaHandler()` for 14+ months and Zappa _feels_ faster.

Perhaps adding a configurable setting is the most conservative approach for now? The default behavior could be off, but allow users to enabled it if desired. Over time we could do more testing and change the default behavior.

This PR would need a quick documentation update - perhaps something like the following:

> ## Cold Starts (Experimental)
> Lambda may provide additional resources than provisioned during cold start initialization. Set `INSTANTIATE_LAMBDA_HANDLER_ON_IMPORT=True` to instantiate the lambda handler on import. This is an experimental feature - if startup time is critical, look into using [Provisioned Concurrency](https://aws.amazon.com/blogs/compute/new-for-aws-lambda-predictable-start-up-times-with-provisioned-concurrency/).

gitumarkk:
This change greatly reduces the Lambda cold start time in a project I am working on as well, by about 6 seconds. A configurable setting should work.

xncbf:
I just edited code by request. plz check
@hellno @wrboyce @javulticat

wrboyce (approved):
Thanks for this!

chiangf:
@wrboyce I noticed an UnboundLocalError when running with INSTANTIATE_LAMBDA_HANDLER_ON_IMPORT=True, so put up https://github.com/zappa/Zappa/pull/1096 as a potential fix

PR #992, opened 2021-06-23
wait for lambda to become active during deploy/update

wrboyce:
Many parts of the deployment process (e.g., scheduling certain events, setting up alb target groups) will fail if the lambda function is not in a ready state, this pr attempts to alleviate that by using a boto waiter during the lambda creation function.

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -1248,7 +1248,7 @@ class Zappa:
             )
 
         # Wait for lambda to become active, otherwise many operations will fail
-        self.wait_until_lambda_function_is_ready(function_name)
+        self.wait_until_lambda_function_is_active(function_name)
 
         return resource_arn
 
@@ -1370,7 +1370,7 @@ class Zappa:
             layers = []
 
         # Wait until function is ready, otherwise expected keys will be missing from 'lambda_aws_config'.
-        self.wait_until_lambda_function_is_ready(function_name)
+        self.wait_until_lambda_function_is_updated(function_name)
 
         # Check if there are any remote aws lambda env vars so they don't get trashed.
         # https://github.com/Miserlou/Zappa/issues/987,  Related: https://github.com/Miserlou/Zappa/issues/765
@@ -1489,18 +1489,23 @@ class Zappa:
 
         return response["FunctionArn"]
 
-    def wait_until_lambda_function_is_ready(self, function_name):
+    def wait_until_lambda_function_is_active(self, function_name):
         """
-        Wait until lambda State=Active and LastUpdateStatus=Successful
+        Wait until lambda State=Active
         """
         # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#waiters
-        kwargs = {"FunctionName": function_name}
         waiter = self.lambda_client.get_waiter("function_active")
         print(f"Waiting for lambda function [{function_name}] to become active...")
-        waiter.wait(**kwargs)
+        waiter.wait(FunctionName=function_name)
+
+    def wait_until_lambda_function_is_updated(self, function_name):
+        """
+        Wait until lambda LastUpdateStatus=Successful
+        """
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#waiters
         waiter = self.lambda_client.get_waiter("function_updated")
         print(f"Waiting for lambda function [{function_name}] to be updated...")
-        waiter.wait(**kwargs)
+        waiter.wait(FunctionName=function_name)
 
     def get_lambda_function(self, function_name):
         """
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -1369,6 +1369,9 @@ class Zappa:
         if not layers:
             layers = []
 
+        # Wait until function is ready, otherwise expected keys will be missing from 'lambda_aws_config'.
+        self.wait_until_lambda_function_is_ready(function_name)
+
         # Check if there are any remote aws lambda env vars so they don't get trashed.
         # https://github.com/Miserlou/Zappa/issues/987,  Related: https://github.com/Miserlou/Zappa/issues/765
         lambda_aws_config = self.lambda_client.get_function_configuration(
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -1336,6 +1336,8 @@ class Zappa:
                     FunctionName=function_name, Qualifier=version
                 )
 
+        self.wait_until_lambda_function_is_updated(function_name)
+
         return resource_arn
 
     def update_lambda_configuration(
@@ -1352,6 +1354,7 @@ class Zappa:
         aws_environment_variables=None,
         aws_kms_key_arn=None,
         layers=None,
+        wait=True,
     ):
         """
         Given an existing function ARN, update the configuration variables.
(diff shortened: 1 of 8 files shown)


pfcodes:
+1 I can't deploy with an ALB because of this..

zappa/core.py
```diff
@@ -1247,6 +1247,10 @@ def create_lambda_function(
                 ReservedConcurrentExecutions=concurrency,
             )
 
+        # Wait for lambda to become active, otherwise many operations will fail
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#waiters
```
ian-whitestone, line 1251: Should we move this reference into the `wait_until_lambda_function_is_ready` docstring?

zappa/core.py
```diff
@@ -1483,34 +1487,18 @@ def rollback_lambda_function_version(
...
-        )
-
-    def wait_until_lambda_function_is_ready(self, function_name):
+    def wait_until_lambda_function_is_ready(self, lambda_arn):
         """
         Continuously check if a lambda function is active.
         For functions deployed with a docker image instead of a
         ZIP package, the function can take a few seconds longer
         to be created or update, so we must wait before running any status
         checks against the function.
```
ian-whitestone, line 1496: May be worth updating this docstring? Originally I added it because we needed to wait until the function was ready for docker deployments, but it sounds like there are other use cases so maybe we should make it more general?

zappa/core.py
```diff
@@ -1247,6 +1247,10 @@ def create_lambda_function(
                 ReservedConcurrentExecutions=concurrency,
             )
 
+        # Wait for lambda to become active, otherwise many operations will fail
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#waiters
+        self.wait_until_lambda_function_is_ready(resource_arn)
+
         return resource_arn
 
     def update_lambda_function(
```
ian-whitestone, line 1256: Should we also add wait until the function is ready when we update the lambda function?

wrboyce: I _think_ the function will always be in Active state after the initial deployment, so this would effectively be a no-op.

ian-whitestone (commented):
Thanks for the PR, this is much cleaner.

Have you been able to test it out with a real deployment? You could leverage [this](https://github.com/ian-whitestone/zappa-serverless-docker/tree/basic_zappa_test) if you don't have an existing setup to test against

wrboyce:
Yeah, I've been using this for a few weeks now over which time I've probably done a good 10+ deploys. No issues thus far although I am using ALB and have APIGateway disabled.

zappa/core.py
```diff
@@ -1483,34 +1487,18 @@ def rollback_lambda_function_version(
...
         return response["FunctionArn"]
 
-    def is_lambda_function_ready(self, function_name):
-        """
-        Checks if a lambda function is active and no updates are in progress.
-        """
-        response = self.lambda_client.get_function(FunctionName=function_name)
-        return (
-            response["Configuration"]["State"] == "Active"
-            and response["Configuration"]["LastUpdateStatus"] != "InProgress"
```
ian-whitestone, line 1493: Do we know if the `waiter.wait()` will exactly mimic ^ this behaviour? My memory is a bit fuzzy, but when I first made this I think I just had the `response["Configuration"]["State"] == "Active"` check, and then had to later add the `and response["Configuration"]["LastUpdateStatus"] != "InProgress"` for some reason. I'm wondering if `waiter.wait()` would only check for the first condition and not necessarily the second

zappa/core.py
ian-whitestone: kk good to leave as is then. easy to add in the future if we realize it's needed

zappa/core.py
wrboyce: Nice catch, I think you are correct in that the waiter only checks the State.

I've just checked [the source](https://github.com/boto/botocore/blob/23ee17f5446c78167ff442302471f9928c3b4b7c/botocore/data/lambda/2015-03-31/waiters-2.json#L21-L46) and it looks like we'd need a `function_updated` waiter too.

illing2005:
What's the status of the PR?
I really need it :pray:

wrboyce:
I need to make the above fixes, it’s been on my mind so will try to get it sorted soon but the real blocker is the CI which is defunct right now, nothing can merge until we get that sorted imo.

wrboyce:
…I wonder if it is worth looking at moving the CI to GHA? Travis sucks 😆

zappa/core.py
javulticat: Waiting for the function to be ready is needed after updating the function as well - this is captured in issues such as #1035

wrboyce: @javulticat I thought you were dealing with that in #1042 but I will make the changes to this PR if you'd prefer.

zappa/core.py
```diff
@@ -1483,34 +1486,18 @@ def rollback_lambda_function_version(
...
-                print("Waiting until lambda function is ready.")
-                show_waiting_message = False
-
-            time.sleep(1)
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#waiters
+        kwargs = dict(FunctionName="{}:{}".format(lambda_arn, ALB_LAMBDA_ALIAS))
+        waiter = self.lambda_client.get_waiter("function_active")
+        print(f"Waiting for lambda function [{lambda_arn}] to become active...")
+        waiter.wait(**kwargs)
+        waiter = self.lambda_client.get_waiter("function_updated")
```
javulticat, line 1498: From my reading of the [AWS docs](https://docs.aws.amazon.com/lambda/latest/dg/functions-states.html), it looks like when creating a function, we'd only need to check the `function_active` waiter and when updating a function, we'd only need to check the `function_updated` waiter.  Would it make sense then to split these up so that we are only doing the necessary check in each case rather than always doing both?

zappa/core.py
javulticat: Happy to consolidate in this PR if you don't mind 🙂

javulticat: Happy to consolidate in this PR if you don't mind 🙂

zappa/core.py
```diff
@@ -1483,34 +1486,18 @@ def rollback_lambda_function_version(
...
+        print(f"Waiting for lambda function [{lambda_arn}] to be updated...")
+        waiter.wait(**kwargs)
```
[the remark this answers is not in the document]
wrboyce: I don't believe this is a regression -- the failure path was never being explicitly handled, although the behaviour will have changed now. A waiter will eventually raise an exception, whereas the previous code would wait forever.

zappa/core.py
wrboyce: @javulticat done, this PR is now checking lambda state prior to updating too.

ian-whitestone (dismissed):

klorophyl (approved):

zappa/core.py
```diff
@@ -1366,6 +1369,9 @@ def update_lambda_configuration(
         if not layers:
             layers = []
 
+        # Wait until function is ready, otherwise expected keys will be missing from 'lambda_aws_config'.
+        self.wait_until_lambda_function_is_updated(function_name)
```
javulticat, line 1373: Thinking about this more, I wonder if this might better live at the end of `update_lambda_function` just like the other waiting function is living at the end of `create_lambda_function` - that might be a safer way to ensure we are waiting for the function to be ready in all cases after create/update.  LGTMO

antoinepay (dismissed):

javulticat (approved):
LGTM

huynhvantrung:
I'm using zappa2 and I get the same error:

```
Traceback (most recent call last):
  File "/root/repo/venv/lib/python3.8/site-packages/zappa/cli.py", line 3422, in handle
    sys.exit(cli.handle())
  File "/root/repo/venv/lib/python3.8/site-packages/zappa/cli.py", line 588, in handle
    self.dispatch_command(self.command, stage)
  File "/root/repo/venv/lib/python3.8/site-packages/zappa/cli.py", line 641, in dispatch_command
    self.update(
  File "/root/repo/venv/lib/python3.8/site-packages/zappa/cli.py", line 1165, in update
    self.lambda_arn = self.zappa.update_lambda_configuration(
  File "/root/repo/venv/lib/python3.8/site-packages/zappa/core.py", line 1395, in update_lambda_configuration
    if lambda_aws_config["PackageType"] != "Image":
KeyError: 'PackageType'
```

Can we fix this for zappa2? Thanks!

PR #993, opened 2021-06-24
relax stage name restrictions when not using apigateway

wrboyce:
Stage name restrictions are in place due to limitations on the naming of CloudFormation stacks which is not a concern when API Gateway is not in use.

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -596,7 +596,7 @@ class ZappaCLI:
         Given a command to execute and stage,
         execute that command.
         """
-
+        self.check_stage_name(stage)
         self.api_stage = stage
 
         if command not in ["status", "manage"]:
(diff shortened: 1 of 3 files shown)


zappa/cli.py
```diff
@@ -598,6 +598,14 @@ def dispatch_command(self, command, stage):
         """
 
         self.api_stage = stage
+        try:
```
hellno, line 601: two minor comments:
- `check_stage_name` throws ` ValueError("AWS requires stage name to match a-zA-Z0-9_")`. You catch it here in `dispatch_command` and raise the same one with slightly more explanation. Could `check_stage_name` throw the long version (with "xyz does not" at the end) and we can avoid catching-and-raising it here?
- suuuper small detail, ignore if you want :) we can run `check_stage_name` before assigning it to `self.api_stage`. Looks a bit cleaner to me to validate before and the using it in the method

hellno (commented):
looks great! 🥳 added a small comment, hopefully not too nitpicky 😅

zappa/cli.py
wrboyce: Hmm, I've got no idea what I was thinking re-raising that Error, you're absolutely correct. Agreed on the second point, too.

wrboyce: Ah, looking at the diff I think I was just moving code around… Fixed now, and looks much tidier!

hellno (approved):
🙌🏽🙌🏽🙌🏽

Issue #998, opened 2021-07-06
Error when deploying AWS Lambda with Zappa package


litalyair:
<html>
<body>

The Error:

Traceback (most recent call last):
--
286 | File "/codebuild/output/src541499700/src/.python/bin/zappa", line 5, in <module>
287 | from zappa.cli import handle
288 | File "/codebuild/output/src541499700/src/.python/lib/python3.8/site-packages/zappa/cli.py", line 44, in <module>
289 | from .core import Zappa, logger, API_GATEWAY_REGIONS
290 | File "/codebuild/output/src541499700/src/.python/lib/python3.8/site-packages/zappa/core.py", line 33, in <module>
291 | import troposphere
292 | File "/codebuild/output/src541499700/src/.python/lib/python3.8/site-packages/troposphere/__init__.py", line 11, in <module>
293 | import cfn_flip
294 | ModuleNotFoundError: No module named 'cfn_flip'

</body>
</html>

## Context

When running the command: zappa package dev -o /tmp/lambda-package.zip
we've got the error above without changing anything in our code

## Your Environment

* Zappa version used: 0.52.0
* Operating System and Python version: python 3.8, ubuntu

alonraiz:
adding `cfn_flip` to my project requirements resolved the issue, but I am not sure what dependency caused this change.

povesma:
I seems that https://pypi.org/project/troposphere/ causes the problem: release `3.0.0` does not install `cfn-flip`. If you return manually to `troposphere==2.7.1` - if works.

fptiangco:
Did the same version downgrade for troposphere as @povesma and can confirm that it fixes the issue.

litalyair:
Thanks for investigating it.
We did the same as well and it resolved it but it is not supposed to be a permanent solution.

fptiangco:
Agreed @litalyair, hopefully @povesma's mention of an ongoing [bug report](https://github.com/cloudtools/troposphere/issues/1926) can help permanently address it.

povesma:
BTW, this is not the only issue with `troposphere==3.0.0`: if you install `cfn-flip` manually and use `troposphere==3.0.0` you'll get

```
  File "./project/venv/lib/python3.8/site-packages/zappa/cli.py", line 1135, in update
    self.zappa.create_stack_template(
  File "./project/venv/lib/python3.8/site-packages/zappa/core.py", line 2375, in create_stack_template
    self.cf_template.add_description("Automatically generated with Zappa")
[...]

PR #999, opened 2021-07-06
Template: add_description -> set_description (stratosphere 3.x).

povesma:
## Description
As `stratosphere` has been upgraded to `3.0.1` , `add_description` is deprecated and removed. `set_description` should be used instead.

## GitHub Issues

https://github.com/zappa/Zappa/issues/998

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -2414,7 +2414,7 @@ class Zappa:
 
         # build a fresh template
         self.cf_template = troposphere.Template()
-        self.cf_template.add_description("Automatically generated with Zappa")
+        self.cf_template.set_description("Automatically generated with Zappa")
         self.cf_api_resources = []
         self.cf_parameters = {}
 


zappa/core.py
```diff
@@ -2414,7 +2414,7 @@ def create_stack_template(
 
         # build a fresh template
         self.cf_template = troposphere.Template()
-        self.cf_template.add_description("Automatically generated with Zappa")
+        self.cf_template.set_description("Automatically generated with Zappa")
```
ian-whitestone, line 2417: Do we also need to update the requirements to force usage of the new troposphere?
If someone is using `troposphere==2.7.0`, will this still work?

Quidge: Yes, I agree that this would break if this PR ran on a deploy that used `troposphere==2.7.0`. The core issue imo is that `troposphere` isn't pinned to a version yet Zappa is relying on a version specific api from troposphere.

@povesma could add a change to the `requirements.txt` that pins the version of `troposphere` to a 3x major version?

shipskin: if we want to keep backwards compatibility why not do something like:
```
if hasattr(self.cf_template, 'add_description'):
    add_description
elif hasattr(self.cf_template, 'set_description'):
    set_description
```

markpeek: The above backwards compatibility change is not needed for troposphere v2.7.x as the change (from v2.4.2 2019-02-02) included both methods (with a deprecation warning on the set_* variant). Of course, the reason for v3.0.0 was to deprecate Python 2.x support and only support Python 3+ in case that factors into zappa's Python version support.

mcrowson:
#1029 also pins to troposphere version, using that one instead

PR #1007, opened 2021-07-17
Fix handling of gzip-encoded text response

suriya:
## Description

When Zappa receives a compressed text/plain response from the
application, it tries to process it as a text response. Instead, Zappa
should treat the response as if it were a binary one and base-64 encode
the response body.

## Github Issues

See issue #2080 binary_support logic in handler.py (0.51.0) broke compressed text response
https://github.com/Miserlou/Zappa/issues/2080

### DIFF
--- a/zappa/handler.py
+++ b/zappa/handler.py
@@ -583,7 +583,16 @@ class LambdaHandler:
                         )
 
                     if response.data:
-                        if (
+                        if settings.BINARY_SUPPORT and response.headers.get(
+                            "Content-Encoding"
+                        ):
+                            # We could have a text response that's gzip
+                            # encoded. Therefore, we base-64 encode it.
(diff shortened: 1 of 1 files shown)


john-parton:
Just need to look at the 'Content-Encoding' header on the werkzeug response. If it's present, then just base64 encode the response as normal.

This logic here:

```python
                        if (
                            settings.BINARY_SUPPORT
                            and not response.mimetype.startswith("text/")
                            and response.mimetype != "application/json"
                        ):
```

Replace it with this

```python
                        if (
                            settings.BINARY_SUPPORT
                            and not response.mimetype.startswith("text/")
                            and response.mimetype != "application/json"
                            and not response.headers.get("Content-Encoding")
                        ):
```

monkut:
Issue supported in:
https://github.com/zappa/Zappa/pull/1155

closing.

PR #1008, opened 2021-07-18
Skip loading credentials for save-python-settings-file

suriya:
## Description

Loading AWS Credentials and creating AWS Client classes take a lot of
time. We do not need to do that for the save-python-settings-file
command.

This is a minor change


monkut (approved):
for the `in` comparison, a list is not required here.  I would use a tuple in this case, but not a major issue.

hellno (approved):
LGTM

PR #1013, opened 2021-07-20
add_description() not exist on Template

mw-ferretti:
## Description
Error trying to deploy Django on AWS lambda. (add_description() does not exist on Template").
I removed this line, and added the description to the Template constructor.
I'm using Django 3.2.5 and Zappa 0.53.0.

## GitHub Issues
I didn't find discussions in the forums.


shipskin:
This would be fixed if we update stratosphere as seen here https://github.com/zappa/Zappa/pull/999/files

Issue #872, opened 2021-02-20
[Migrated] Make S3 triggers apply correctly with multiple stages using the same bucket


jneves:
Originally from: https://github.com/Miserlou/Zappa/issues/2126 by [arcadecoffee](https://github.com/arcadecoffee)

## Description
The S3EventSource handler from Kappa has been re-implemented to allow more precise addition and removal of S3 bucket event triggers allowing Zappa to work when multiple stages, app_functions, or even whole projects to apply and manage their own triggers on S3 buckets.

## GitHub Issues
[S3 triggers apply incorrectly with multiple stages using the same bucket #2111 ](https://github.com/Miserlou/Zappa/issues/2111)

PR #1025, opened 2021-08-16
Make S3 triggers apply correctly with multiple stages using the same bucket

kimdorr:
## Description
The S3EventSource handler from Kappa has been re-implemented to allow more precise addition and removal of S3 bucket event triggers allowing Zappa to work when multiple stages, app_functions, or even whole projects to apply and manage their own triggers on S3 buckets.

(This is an updated version of https://github.com/Miserlou/Zappa/pull/2126, originally opened by @arcadecoffee)

## GitHub Issues
fixes https://github.com/zappa/Zappa/issues/872


monkut:
zappa is currently tightly integrated with kappa, this may be a good starting point to migrate away from kappa.

Issue #1027, opened 2021-08-20
Feature: Automatic gzipping of responses


john-parton:
I had a few large pages which exceeded lamba's response size limits, so I patched Zappa to automatically compress responses with gzip.

## Context
My app is behind CloudFront, so I didn't do any checking for `Accept-Encoding` and always gzip the response if the base64 encoded gzip'd response is smaller than the non-compressed response.

HTML and JSON are _extremely_ compressible, so even with the overhead of base64 encoding, you often can squeeze a ton more data in.

monkut:
This may be a good *option* to have in zappa, but currently external libraries, like [whitenoise](http://whitenoise.evans.io/en/latest/), can perform this for you.

john-parton:
I'm not sure you're correct. Whitenoise is just for static files. 

My (closed) PR gzip'd every response, including those dynamically generated by your views.

I'm not a Zappa user anymore, so I probably will not be contributing any further changes.

PR #1028, opened 2021-08-20
Automatically Gzip Responses

john-parton:
Automatically gzip responses if the base64 encoded response is smaller.

Issue: https://github.com/zappa/Zappa/issues/1027

I have this deployed in production and it works well.

This will fail if a client doesn't support gzip compression. This isn't an issue if your app is behind CloudFront.

### DIFF
--- a/zappa/handler.py
+++ b/zappa/handler.py
@@ -31,6 +31,29 @@ logging.basicConfig()
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
+import gzip
+
+# Django uses 200 bytes, but we're going to use 2k
+MIN_GZIP_LENGTH = 2_048
+
+# Return (body, content_length) tuple
+def _get_gzipped_body(response):
+    # Not sure this should ever be there?
+    if response.headers.get("Content-Encoding"):
+        print("Already have Content-Encoding in response")
+        return None, None
+
+    # Don't waste time
+    if len(response.data) < MIN_GZIP_LENGTH:
+        # print("Small response")
+        return None, None
+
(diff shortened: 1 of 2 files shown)


ghostmaster-ai:
Pease make the same pull request on  [Zappa2](https://github.com/ghostmaster-ai/Zappa2), which is a branch of this repo. Since This repo has been deprecated I will take up [maintenance.](https://github.com/ghostmaster-ai/Zappa2)
[PyPi](https://pypi.python.org/pypi/zappa2) - pip install zappa2

zappa/handler.py
```diff
@@ -593,19 +621,39 @@ def handler(self, event, context):
                             ).decode("utf-8")
                             zappa_returndict["isBase64Encoded"] = True
                         else:
-                            zappa_returndict["body"] = response.get_data(as_text=True)
+
+                            gzipped_body, gzipped_content_length = _get_gzipped_body(response)
```
alhoo, line 625: It would be neat if you would verify that the `headers["Accept-Encoding"]` contains "gzip" before doing the compression. This would make this feature work according to the current http specs. See

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

alhoo: Or maybe use https://github.com/colour-science/flask-compress instead of  implementing it your self?

monkut:
A major update to how BINARY data is used was brought in recently.
Since there are a number of conflicts here, I think this PR would need to be redone.

I'm going to close this PR and move discussion to the related issue.

PR #1040, opened 2021-09-16
fix https://github.com/zappa/Zappa/issues/1039 prohibit public access…

bruceduhamel:
… to created lambda functions by setting SourceAccount when creating lambda resource policy.

## Description
 
When granting permissions to invoke a lambda for an event source such as s3 object created, specify the SourceAccount to restrict access to the lambda function to strictly that account. This is a minor change to prevent public access to a lambda function. 

## GitHub Issues

https://github.com/zappa/Zappa/issues/1039

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,3 @@
-<p align="center">
-  <img src="http://i.imgur.com/oePnHJn.jpg" alt="Zappa Rocks!"/>
-</p>
 ## DEPRECATED - This project is no longer maintained
 The maintainers do not have the time to keep this alive. Feel free to fork and have a go at it if you have the time and inclination.
 


ghostmaster-ai:
Pease make the same pull request on  [Zappa2](https://github.com/ghostmaster-ai/Zappa2), which is a branch of this repo. Since This repo has been deprecated I will take up [maintenance.](https://github.com/ghostmaster-ai/Zappa2)
[PyPi](https://pypi.python.org/pypi/zappa2) - pip install zappa2

bruceduhamel:
@ghostmaster-ai any chance you'll be able to publish the package to pypi as zappa, rather than zappa2?

ghostmaster-ai:
> @ghostmaster-ai any chance you'll be able to publish the package to pypi as zappa, rather than zappa2?

for that to happen the owners/maintainers of PyPi account will have delete their project on PyPi since there cannot be 2 projects with the same name. Not only do I not see that happening I don't think it's a good idea to take down the original project as they might want to take up maintenance again in the future, till then i hope [Zappa2](https://pypi.org/project/zappa2/) does the trick.

PR #1050, opened 2021-10-01
Fix encoding error that occurs when reading readme

xncbf:
If you use a language other than English in Windows environment (2 bytes languages), cp949 error occurs during installation

```
[pipenv.exceptions.InstallError]:     Complete output (5 lines):
[pipenv.exceptions.InstallError]:     Traceback (most recent call last):
[pipenv.exceptions.InstallError]:       File "<string>", line 1, in <module>
[pipenv.exceptions.InstallError]:       File "C:\Users\AppData\Local\Temp\pip-install-soudjjra\zappa_0b8c2fb4266844e8bbe754fa82cfb4f2\setup.py", line 8, in <module>
[pipenv.exceptions.InstallError]:         long_description = readme_file.read()
[pipenv.exceptions.InstallError]:     UnicodeDecodeError: 'cp949' codec can't decode byte 0xe3 in position 73845: illegal multibyte sequence
```


javulticat (approved):
LGTM - thanks!

PR #1052, opened 2021-10-02
Add exclude_glob documentation

tomdaly:
## Description

Documentation contains incorrect information about `exclude` setting and no information about `exclude_glob`

### Changes
- Add `exclude_glob` documentation for `zappa_settings.json` to README
- Update `exclude` documentation example to mention regex, not glob

## GitHub Issues
Resolves https://github.com/zappa/Zappa/issues/422


wrboyce (approved):

monkut (approved):

willrogers:
This is still incorrect! The `exclude` pattern are not regex, they are wildcards as described in the [fnmatch module](https://docs.python.org/3/library/fnmatch.html).

The example `tests/` won't be excluded. It would if it was `tests` or if `tests/` was in the `exclude_glob` array.

PR #1060, opened 2021-10-15
remove library that aws lambda cannot find

jottenlips:
## Description
 

I do not see why this hardcoded library is necessary in the source code. If someone wants to include it they can add it to `include: []`. 
We ran into this issue when updating from an old version of Zappa that did not have this hardcoded reference. We are also using MySQL. Happy to discuss alternatives, this just seemed the most straightforward solution. Thanks!

## GitHub Issues

https://github.com/zappa/Zappa/issues/940


jottenlips:
I could also see Zappa forking https://github.com/Miserlou/lambda-packages, seems like there is a lot of good stuff in it like the ability to use Pillow on lambda

hellno:
Thanks for bringing this issue up and giving context, super helpful 👍🏽

Will this change break existing projects using mysql based on zappa?
If yes, we need clear instructions how users can manually add it back via `include: [...]`

jottenlips:
Current projects using mysql broke when lambda-packages was removed. This change should have no effect. To use mysql with zappa as of now you have to either make your own lambda layer or rely on pymysql.

javulticat:
Looks like `black` needs to be run.  Try running `make black` before pushing.  LGTMO

jottenlips:
Thanks!

jottenlips:
looks like the automation wants this to be reviewed again? 
"At least 1 approving review is required by reviewers with write access."

jottenlips:
thanks @javulticat! glad to see the tests are passing

javulticat (approved):

PR #1065, opened 2021-11-03
Increase lambda client read timeout to 15 minutes

### LINKED ISSUE #1064: Lambda client read timeout should match maximum Lambda execution time
## Context
The boto3 Lambda client used by the Zappa CLI has its read timeout set to [5 minutes](https://github.com/zappa/Zappa/blob/master/zappa/core.py#L340), while the maximum Lambda function execution time is now [15 minutes](https://aws.amazon.com/about-aws/whats-new/2018/10/aws-lambda-supports-functions-that-can-run-up-to-15-minutes). This means that `zappa invoke` or `zappa manage` could time out before the actual function execution completes. I encountered this recently with a long-running Django management command, using Zappa 0.54 on Python 3.8.

## Expected Behavior

The Lambda client used by the CLI should not time out before the execution completes.

## Actual Behavior

The client times out after 5 minutes regardless of the configured execution time, which can be up to 15 minutes.

## Possible Fix

The timeout should be updated to the maximum Lambda function execution time.

## Steps to Reproduce

1. Set up a Lambda function with a timeout greater than 5 minutes.
2. Deploy some code that takes more than 5 minutes to run.
3. Invoke the long-running code using `zappa invoke` or `zappa manage`.

## Your Environment

* Zappa version used: 0.54
* Operating System and Python version: macOS 11.6.1, Python 3.8.11
* The output of `pip freeze`: `zappa==0.54.0`

rolandcrosby-check:
## Description
 
This change increases the read timeout on the Zappa CLI's Lambda client to 15 minutes.

As of 2018, Lambda functions [can be configured][1] to run for up to 15 minutes per execution, but the configuration for the Zappa CLI's Lambda client had its timeout set to 5 minutes. This meant that the Lambda client would time out waiting for a response from a long-running invocation (e.g. a Django management command that runs for more than 5 minutes) while the actual function could continue to execute for up to 15 minutes, depending on its configuration.

I believe the comment in the README regarding the 30-second API gateway timeout's effect on long-running management command invocations is incorrect or misleading. Django management commands invoked via `zappa manage` execute the Lambda function directly using a boto3 Lambda client (in the `Zappa.invoke_lambda_function` method), so the API Gateway timeout should not affect these invocations at all.

[1]: https://aws.amazon.com/about-aws/whats-new/2018/10/aws-lambda-supports-functions-that-can-run-up-to-15-minutes
## GitHub Issues

Fixes #1064.

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,8 +420,6 @@ Commands which require direct user input, such as `createsuperuser`, should be [
 
 For more Django integration, take a look at the [zappa-django-utils](https://github.com/Miserlou/zappa-django-utils) project.
 
-_(Please note that commands which take over 30 seconds to execute may time-out preventing output from being returned - but the command may continue to run. See [this related issue](https://github.com/Miserlou/Zappa/issues/205#issuecomment-236391248) for a work-around.)_
-
 ### SSL Certification
 
 Zappa can be deployed to custom domain names and subdomains with custom SSL certificates, Let's Encrypt certificates, and [AWS Certificate Manager](https://aws.amazon.com/certificate-manager/) (ACM) certificates.
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -331,13 +331,14 @@ class Zappa:
         # Some common invocations, such as DB migrations,
         # can take longer than the default.
 
-        # Note that this is set to 300s, but if connected to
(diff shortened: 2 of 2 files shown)


javulticat (approved):
LGTM

PR #1069, opened 2021-11-08
Decode `zappa invoke` output regardless of `--no-color` option

### LINKED ISSUE #1068: `zappa invoke --no-color` should decode result before printing
Currently, when using the `zappa invoke` or `zappa manage` CLI tools with the `--no-color` argument, Zappa directly prints the result of base64-decoding the [`LogResult`](https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html#API_Invoke_ResponseSyntax) attribute from the call to the Lambda `Invoke` action. This displays the output as a Python bytes literal (`b'like this'`) with all newlines and quotation marks escaped. While this does allow for representation of sequences that are not valid UTF-8, this is not a particularly useful format for processing in other downstream tools, and it's not a side effect that I expected just from specifying that I don't want ANSI color sequences in the output.

(While we're looking at this, it would also be nice to make `zappa invoke` match the default behavior of `click` and `zappa tail` by disabling colorization if output is not being written to a terminal.)

## Context

In [zappa/cli.py](https://github.com/zappa/Zappa/blob/679e98eb7c38913022a5cb0d0610e4529d4340d2/zappa/cli.py#L1538-L1545), the `--no-color` CLI argument is checked before printing the response received from Lambda. If it's present, Zappa base64-decodes the result and prints that directly. Otherwise, Zappa base64-decodes the result, decodes the contents of the result into a UTF-8 string, and colorizes that result:

```python
if "LogResult" in response:
    if no_color:
        print(base64.b64decode(response["LogResult"]))
    else:
        decoded = base64.b64decode(response["LogResult"]).decode()
        formatted = self.format_invoke_command(decoded)
        colorized = self.colorize_invoke_command(formatted)
        print(colorized)
```

## Expected Behavior

Zappa should decode the response as UTF-8 regardless of whether the output is being colorized. (If the result can't be decoded, it would be nice to fall back to printing the result as bytes.)

## Actual Behavior

Specifying `--no-color` causes Zappa to skip decoding the response into a UTF-8 string.

## Possible Fix

Decode the `LogResult` attribute as UTF-8 regardless of whether the output is being colorized.

## Steps to Reproduce

1. Deploy some code via Zappa
2. Invoke some code using `zappa invoke` or `zappa manage`, and use the `--no-color` argument

## Your Environment

* Zappa version used: 0.54.0
* Operating System and Python version: macOS 11.6.1, Python 3.8.11
* The output of `pip freeze`: `zappa==0.54.0`

rolandcrosby-check:
## Description
 
When a user calls `zappa invoke` or `zappa manage`, always decode the output's `LogResult` as UTF-8, regardless of whether the use has specified `--no-color`. If the output cannot be decoded as UTF-8, fall back to printing the raw bytes as a `b''` literal.

Also disables colorization if the output is not being written to a terminal.

## GitHub Issues

Fixes #1068.


hellno (dismissed):
Thanks for the detailed issue and great PR 🙌🏼🙌🏼 
looks good to me!

rolandcrosby-check:
@hellno TFTR! Let me know if you think this needs a test before you can merge it; I see that coverage went down but I didn't see any existing test that this could fit into neatly.

rolandcrosby-check:
Slightly refactored and added a test so this should no longer cause issues with reduced coverage.

wrboyce (approved):

rolandcrosby-check:
@wrboyce Just resolved the merge conflict; let me know if you need anything else from me to merge this.

PR #1071, opened 2021-11-09
Don't install dependencies

tams:
Allows the project using Zappa to declare Zappa dependencies themselves to avoid getting conflicts with their own deps.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,18 +7,10 @@ from zappa import __version__
 with open('README.md') as readme_file:
     long_description = readme_file.read()
 
-with open(os.path.join(os.path.dirname(__file__), 'requirements.in')) as f:
-    required = f.read().splitlines()
-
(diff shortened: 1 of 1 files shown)


tams:
Did not intend to upstream this. Please ignore.

tams:
Anyway, since I already fat fingered this PR, it should be of some help in case anyone lands on a similar error where Zappa's injected dependencies collide with your own stuff:

```
ERROR: Cannot install pip-tools==6.4.0 and pip==21.0.1 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

You can cut out Zappa's own dependencies like in this PR and instead add them as your own project's dependencies with versions that play nice.

PR #1075, opened 2021-11-11
automatically publish tagged versions to pypi

wrboyce:


.github/workflows/cd.yaml
```diff
@@ -0,0 +1,24 @@
+name: CD
+
+on:
+  push:
+    tags: ["v*"]
```
javulticat, line 5: Our release tags so far have not started with `v`, see: https://github.com/zappa/Zappa/tags.  Perhaps we could match on semver regex instead?  `\d+\.\d+\.\d+`

Per the [docs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet), it seems the GHA equivalent would be: `[0-9]+.[0-9]+.[0-9]+`

.github/workflows/cd.yaml
```diff
@@ -0,0 +1,24 @@
+name: CD
+
+on:
+  push:
```
javulticat, line 4: Should this be a `create` trigger instead?   A `pypi` release should be immutable to avoid unexpected breaking changes for our users, so we don't want to republish a tag that we've already published to `pypi`.  In case a tag accidentally gets overwritten or something.

.github/workflows/cd.yaml
```diff
@@ -0,0 +1,24 @@
+name: CD
```
javulticat, line 1: Nit: Should this be `publish` or `package` or something like that?  I wouldn't really call this continuous or a deployment 😆

.github/workflows/cd.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Install `pypa/build`
+        run: python -m pip install build
+      - name: Build sdist and wheel
+        run: python -m build --sdist --wheel --outdir ./dist/
+      - name: Publish package to PyPI
+        uses: pypa/gh-action-pypi-publish@master
```
javulticat, line 22: Let's use the latest stable release instead of the tip of `master`, as recommended in the [docs](https://github.com/pypa/gh-action-pypi-publish/tree/master#usage):

`pypa/gh-action-pypi-publish@release/v1`

javulticat (commented):
Very excited for this! 🚀

.github/workflows/cd.yaml
wrboyce: Nice catch

.github/workflows/cd.yaml
wrboyce: If we had access to the testpypi Zappa package it would be somewhat more continuous at least 😬

.github/workflows/cd.yaml
wrboyce: I meant to bring this up on Slack, as it does seem the convention [changed at some point](https://github.com/Miserlou/Zappa/tags).

wrboyce: I've at least changed it to the more explicit `v[0-9]+.[0-9]+.[0-9]+` for now.

javulticat: Should we support both to be safe?  I think just adding a `?` after the `v` should do it:  `v?[0-9]+.[0-9]+.[0-9]+`

hellno:
Thanks for adding this @wrboyce 💪🏼 great work (and sorry I misclicked on merge-commits from master 😓)

.github/workflows/cd.yaml
wrboyce: good idea, done 👍

.github/workflows/cd.yaml
```diff
@@ -0,0 +1,24 @@
...
+on:  # yamllint disable-line rule:truthy
+  create:
+    tags: ["v?[0-9]+.[0-9]+.[0-9]+"]
+
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout Code Repository
+        uses: actions/checkout@v2
```
javulticat, line 12: Quick question: Is this checking out the tag being created?  Or is it checking out `master`?  One possible issue with checking out `master` is that we could create a new tag and then immediately afterward merge another change into `master` and if this command is checking out `master`, it would include an extra commit not found in the tag.

wrboyce: checkout would get the commit triggering the job, in this case the tagged release.

MRigal (approved):
LGTM!

javulticat (approved):

PR #1077, opened 2021-11-12
Convert project to using 'pipenv' to manage dependencies

javulticat:
## Description
Converts the project to use [`pipenv`](https://pipenv-fork.readthedocs.io/en/latest/index.html) to manage dependencies, as discussed in Slack.


Pipfile
```diff
@@ -0,0 +1,48 @@
...
+django-stubs = "*"
+flake8 = "*"
+Flask = "*"
+isort = "*"
+mock = "*"
+mypy = "*"
+nose = "*"
+nose-timer = "*"
+pipenv = "*"
+placebo = "<0.10"
```
hellno, line 20: duplicate of requirement in [packages] below

hellno (approved):
love this! thanks for the PR 💪🏼 much easier to maintain this way (at least to my mind)

I assume you didn't commit the `Pipfile.lock` on purpose. Is it not needed in this case? (From other projects I'm used to adding it, but that might not be the best way to go 😊)

Basically my question is: are the dependencies fixed enough without having `Pipfile.lock`?

Pipfile
javulticat: Yep good call - this was a little confusing because `placebo` _should_ live in `dev-packages`, but we need it to live in `packages` temporarily until we're able to unpin its version.  I removed it here and left a comment by the entry in `packages` to remind us to move it to `dev-packages` once the version is unpinned.

javulticat:
> love this! thanks for the PR 💪🏼 much easier to maintain this way (at least to my mind)
> 
> I assume you didn't commit the `Pipfile.lock` on purpose. Is it not needed in this case? (From other projects I'm used to adding it, but that might not be the best way to go 😊)
> 
> Basically my question is: are the dependencies fixed enough without having `Pipfile.lock`?

@hellno, I didn't commit `Pipfile.lock` in this case because the lock file is specific to a single version of Python, whereas Zappa supports Py 3.6 - 3.9.  Our options would seem to be either:

1. Commit 1 lock file per version of Python we support
2. Don't commit any lock file, and rely on having dependencies pinned/bounded appropriately in `Pipfile` (this is the approach I went with as it minimized clutter in the repo)

hellno:
Love it! The many python versions + less clutter makes a lot of sense 👍🏼

Issue #1073, opened 2021-11-10
Use Route53 Alias DNS records rather than CNAME


yapret:
## Context

When setting up a custom domain hosted on Route53, Zappa uses an Alias record when that custom domain is the apex of the Hosted Zone, and a CNAME record otherwise.

To me, CNAME records seem:
- Slower as the client needs to do two requests (one on the custom domain, then one on the CNAME target)
- More expensive as CNAME requests are billed where Alias ones aren't
- Problematic when needing to add other types of records for the same hostname (TXT for SPF, for example)

This behavior cannot be configured at the moment, and adds some complexity to the code.

Relevant part of the code:
https://github.com/zappa/Zappa/blob/679e98eb7c38913022a5cb0d0610e4529d4340d2/zappa/core.py#L2652-L2695

## Expected Behavior

Zappa should create Alias A/AAAA records when adding a custom domain.

## Actual Behavior

Zappa creates either a CNAME or Alias A record depending on whether the hostname is the apex of the Hosted Zone.
## Possible Fix

Make Alias A/AAAA records the standard for all kinds of hostnames.

## Steps to Reproduce
n/a

## Your Environment

* Zappa version used: 0.54.0

javulticat:
Confusingly, `A`/`AAAA` records are "address" records, not "alias" records.  `CNAME` is the closet thing to an "alias" record in vanilla DNS.

But Route53 has a _custom_ "alias" concept which can apply to both types of records.  See: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-choosing-alias-non-alias.html

This provides the benefits you listed, as well as the benefits of `CNAME` records (such as improved portability and geotargeting).

I think the best solution here would be to always make both types of records use a Route53 "alias".  I opened a PR for this here: #1079

PR #1079, opened 2021-11-12
Always use aliased Route53 DNS records

javulticat:
## Description
In Route53, an "aliased" record is a special custom record type that can apply to almost any kind of DNS record.  Let's always use aliased records to increase performance and lower cost.

See: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-choosing-alias-non-alias.html

## GitHub Issues
#1073

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -2659,32 +2659,17 @@ class Zappa:
             self.route53.get_hosted_zone(Id=zone_id)["HostedZone"]["Name"][:-1]
             == domain_name
         )
-        if is_apex:
-            record_set = {
-                "Name": domain_name,
-                "Type": "A",
-                "AliasTarget": {
-                    "HostedZoneId": "Z2FDTNDATAQYW2",  # This is a magic value that means "CloudFront"
-                    "DNSName": dns_name,
-                    "EvaluateTargetHealth": False,
-                },
-            }
-        else:
-            record_set = {
-                "Name": domain_name,
-                "Type": "CNAME",
-                "ResourceRecords": [{"Value": dns_name}],
-                "TTL": 60,
-            }
 
-        # Related: https://github.com/boto/boto3/issues/157
-        # and: http://docs.aws.amazon.com/Route53/latest/APIReference/CreateAliasRRSAPI.html
-        # and policy: https://spin.atomicobject.com/2016/04/28/route-53-hosted-zone-managment/
-        # pure_zone_id = zone_id.split('/hostedzone/')[1]
+        record_set = {
+            "Name": domain_name,
+            "Type": "A" if is_apex else "CNAME",
+            "AliasTarget": {
+                "HostedZoneId": "Z2FDTNDATAQYW2",  # This is a magic value that means "CloudFront"
+                "DNSName": dns_name,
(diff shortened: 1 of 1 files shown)


wrboyce (approved):
Nice

yapret:
Thanks for your PR. I'm worried that this isn't going to work, as Route53 only resolves alias records when their target is of the same type (so, for example, in order for a CNAME alias record to work, the target also needs to be a CNAME record, which isn't the case with CloudFront's A records).

See https://aws.amazon.com/premiumsupport/knowledge-center/route-53-create-alias-records/?nc1=h_ls:

> ## CNAME support
> 
> Route 53 follows the pointer in an alias record **only if the record type matches**. To create an alias record for a CNAME record, the alias target must resolve to a CNAME value.
> 
> Even if you create an invalid CNAME(ALIAS) record, Route 53 can't resolve the IP address of the endpoint. As a result, Route 53 returns an SERVFAIL/REFUSED RCODE error.

and

> Be sure to use the correct DNS record type for IP addresses in the service you're pointing to:
> - Amazon API Gateway custom regional API and edge-optimized API – A record (IPv4)

One way to mitigate this problem would be to completely ditch CNAME records and use only A, as proposed in #1074.

javulticat:
@yapret Thanks- admittedly DNS is not something I'm an expert on. A few questions:
* Ifwe are already using `CNAME` successfully, shouldn't `ALIAS(CNAME)` work too? Isn't the problem with `A` records that you lose geotargeting information?

yapret:
@javulticat No problem!

> * If we are already using `CNAME` successfully, shouldn't `ALIAS(CNAME)` work too?

I don't think so because the record targeted by an alias should be of the exact same type. It's part of Route53's requirements for the alias feature, whereas the standardized CNAME records don't have such a limitation.

> * Isn't the problem with `A` records that you lose geotargeting information?

I imagine that you're referring to the [NS1 blog post](https://ns1.com/blog/the-difference-between-alias-and-cname) about ALIAS records. To my understanding, as ALIAS records aren't part of the DNS standards, every provider's implementation is different, so what applies to NS1 may not hold true for AWS. As far as I know, this problem doesn't exist on Route53, and alias records are in fact the recommended way to use custom domains on API Gateways and, more generally, CloudFront distributions:

> With Route 53 you can create an A record alias for your custom domain name and specify the CloudFront distribution domain name as the alias target. This means that Route 53 can route your custom domain name even if it is a zone apex. For more information, see Choosing Between Alias and Non-Alias Resource Record Sets in the Amazon Route 53 Developer Guide.
> Use of A-record aliases also eliminates exposure of the underlying CloudFront distribution domain name because the domain name mapping takes place solely within Route 53. For these reasons, **we recommend that you use Route 53 A-record alias whenever possible.**

From https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-edge-optimized-custom-domain-name.html#how-to-custom-domains-console

More info: 
* https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-to-api-gateway.html
* https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-choosing-alias-non-alias.html

Issue #854, opened 2021-02-20
[Migrated] Unique rule names for multiple CloudWatch expressions (#1727)


jneves:
Originally from: https://github.com/Miserlou/Zappa/issues/2102 by [edudobay](https://github.com/edudobay)

## Description

A mixture of two cases that were handled individually uncovers an incorrect behavior: when a function has multiple event expressions AND the resulting rule name would be >= 64 characters long, all expressions get the same hash, so only the last expression is scheduled.

In my fix, the index of the expression is used to build the hash, but only if the index is not zero. My rationale for this was preserving hashes for current rules. This seems unneeded (it doesn't seem like it would break anything without caring about this) but may prevent surprises. I'd like some feedback on whether this is an appropriate choice.

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/1727

PR #1080, opened 2021-11-12
Ensure unique scheduled event names

javulticat:
## Description
Ensure event names are unique.  To do so, the process of creating event names was streamlined to make it easier to follow.

## GitHub Issues
#854


edudobay:
see also #2 which solves the same issue.

hellno (approved):
LGTM 👍🏽

PR #1083, opened 2021-11-18
Support Newest Manylinux Version for dependencies

phil-airship:
Zappa was hard coded to only allow Manylinux 1, 2010, and 2014.
Manylinux 2_24 is getting more traction, so dependencies are becoming incompatible with Zappa over time.
More information on new Manylinux versions can be found here - https://github.com/pypa/manylinux

An important example is Cryptography, where version 3.4.7 is the last one to support old-format linux wheels.
https://pypi.org/project/cryptography/3.4.7/#files
vs newest:
https://pypi.org/project/cryptography/#files

Tested on my env to make sure that manylinux 2_24 wheels would work without issue in Lambda


hellno (approved):
Thanks for creating this PR, super helpful (and I wasn't aware of the new manylinux dependency format at all, so thanks for pointing that out)

phil-airship:
Is there a reason it things the coverage decreased for some python version, but stayed the same for overall coverage and py3.7?

hellno:
We need to fix the flaky coverage tests separately, it's not related to this PR.
The +-0.0% coverage changes on different python versions shouldn't fail at all :)

hellno:
Updated from master to test new coveralls threshold, seems to work 👍🏼

hellno:
Thanks for the PR @phil-airship, I went ahead and merged it

Issue #1082, opened 2021-11-15
 Work with Plotly Dash devs to doc Zappa deploy


strongbad03:
x-posted with Dash [here](https://github.com/plotly/dash/issues/1833)

## Context
There's been an lingering issue with using Dash and Zappa producing 403, 404, and 500 errors, depending upon how the app file and Zappa JSON settings are written. It's difficult to deploy a Dash app with Zappa -- a major bummer, since the combination of Dash's slickness with Zappa's serverlessness is a killer peanut-butter-and-chocolate combo.

## Expected Behavior
I'd like a "Hello World" doc for a serverless Dash app using Zappa OR a Dash-side OR Zappa-side patch to address behavior. Currently using latest for each, zappa 0.54.1 and Dash 2.0.0 in Python 3.8. 

## Actual Behavior
I've followed closely the mods [discussed here](https://github.com/plotly/dash/issues/22) and [the sample app](https://github.com/mcrowson/zappa_dash) but am still hitting errors. Tried the mod [here](https://github.com/Miserlou/Zappa/issues/2200#issuecomment-772702958) and am still getting a 404. 

Here's the three different file configs I've tried:

1. Dash object as app_function
If I use this config, a visit to the app URL results in a 500 error `Dash object is not callable`, as [discussed](https://github.com/plotly/dash/issues/22#issue-237662489). [Tinkering with the Dash object](https://github.com/plotly/dash/issues/22#issuecomment-310231167) doesn't address this. 

2. Flask server as app_function
[This suggestion](https://github.com/plotly/dash/issues/22#issuecomment-310516941) results in a 403 when visiting the app URL.

3. Same as 2, plus url_base_pathname. Results in a 404.

## Possible Fix
n/a

## Steps to Reproduce
Details in ticket [x-posted in Dash project](https://github.com/plotly/dash/issues/1833)

## Your Environment

* Zappa version used:
* Operating System and Python version: Ubuntu 20.2, Python 3.8
* The output of `pip freeze`:
```
appdirs==1.4.3
argcomplete==1.12.3
boto3==1.20.5
botocore==1.23.5
Brotli==1.0.9
CacheControl==0.12.6
certifi==2019.11.28
cfn-flip==1.3.0
chardet==3.0.4
click==8.0.3
colorama==0.4.3
contextlib2==0.6.0
dash==2.0.0
dash-bootstrap-components==1.0.0
dash-core-components==2.0.0
dash-html-components==2.0.0
dash-table==5.0.0
[...]

PR #1095, opened 2021-12-22
Add example of dash deployment

alhoo:
There's been an lingering issue with using Dash and Zappa producing 403,
404, and 500 errors, depending upon how the app file and Zappa JSON
settings are written. It's difficult to deploy a Dash app with Zappa --
a major bummer, since the combination of Dash's slickness with Zappa's
serverlessness is a killer peanut-butter-and-chocolate combo.

This commit contains a minimal example of a dash app to help users get
started with deploying dash applications using Zappa.

Closes https://github.com/zappa/Zappa/issues/1082


monkut:
This is an old PR, and samples may be better as a blog post or separate repository.
Closing

PR #1096, opened 2021-12-24
Update handler to be global if INSTANTIATE_LAMBDA_HANDLER_ON_IMPORT=True

chiangf:
INSTANTIATE_LAMBDA_HANDLER_ON_IMPORT was introduced in https://github.com/zappa/Zappa/pull/982 but I'm noticing an error `UnboundLocalError: local variable 'handler' referenced before assignment` if I set the env var, which I believe is because `handler` needs to be global.

### DIFF
--- a/zappa/handler.py
+++ b/zappa/handler.py
@@ -248,9 +248,7 @@ class LambdaHandler:
 
     @classmethod
     def lambda_handler(cls, event, context):  # pragma: no cover
-        global handler
-        if not os.environ.get("INSTANTIATE_LAMBDA_HANDLER_ON_IMPORT"):
-            handler = cls()
+        handler = global_handler or cls()
         exception_handler = handler.settings.EXCEPTION_HANDLER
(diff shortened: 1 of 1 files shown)


zappa/handler.py
```diff
@@ -248,6 +248,7 @@ def import_module_and_get_function(whole_function):
 
     @classmethod
     def lambda_handler(cls, event, context):  # pragma: no cover
+        global handler
```
javulticat, line 251: Typically we'd prefer to avoid using `global`.  Could simply avoiding the name collision do the same - something like this?

```python
global_handler = LambdaHandler() if os.environ.get("INSTANTIATE_LAMBDA_HANDLER_ON_IMPORT") else None

...

    @classmethod
    def lambda_handler(cls, event, context):
        handler = global_handler or cls()
```

javulticat (approved):
Thanks!

PR #1097, opened 2021-12-25
Use Poetry for managing virtualenvs and building

tommie-lie:
## Description
This PR introduces [poetry](https://python-poetry.org/) to the project. Poetry is a Python build tool that manages virtualenvs (separate envs per Python version) and allows for easy building and publishing to PyPI.

The currently used Pipenv has some drawbacks:
* as noted in the PR itself (https://github.com/zappa/Zappa/issues/1077#issuecomment-970640705_), Pipenv's lock file should not be committed when using multiple versions of Python
* Pipenv development has been shaky at best over the last two years and is actively looking for maintainers.
* Pipenv still depends on setuptools and `setup.py`, with the latter being tedious to maintain (notice the missing files in the deleted `MANIFST.in`), a latent security issue (because it can contain arbitrary Python code which packagers are supposed to run on their system) and can really be automated with more modern tools.
* Poetry uses the standard `pyproject.toml` file for managing all project description purposes. This file format has been standardized and will be the future of Python packaging (c.f. https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/). Many other tools (like pytest, black or mypy can be configured using pyproject.toml, so there will be only one configuration file in the future. (I know, this is not a drawback of Pipenv, but an advantage of poetry :-))

Consequently, this PR removes `setup.py` and adds a `pyproject.toml` which contains the same information (I know there are still errors in it that are inherited from the original `setup.py`. I can provide a separate PR once this is merged or amend this PR if need be). Dependency versions are also locked (in `poetry.lock`) across Python versions.

Poetry's usage is quite similar to `npm`, if you are familiar with it. After checking out the sourcecode, you just `poetry install` and get a clean virtualenv with all dependencies installed. To get an interactive shell into this virtualenv, run `poetry shell`. Add new dependencies using `poetry add newdependency`. No need to manually fiddle with pip or venv anymore.

## A note to maintainers:
As mentioned, poetry makes it easier to build and publish new versions of Zappa to PyPI. You can use the fixed Makefile or just run `poetry build` directly to get both, a source and a wheel distribution. In addition, you can run `poetry publish` to directly push a new version to PyPI (after logging in, of course). I don't know your current workflow, but I guess it can't get any easier than this.

## Another note to maintainers:
I removed `__version__` from `zappa/__init__.py`. It was not used anywhere but `setup.py`, which doesn't exist anymore. The canonical way to set a project's version should be `pyproject.toml`. If we ever need the installed version in the sourcecode, the modern way is to use `importlib.metadata` which directly queries Python's installaed package metadata.


wrboyce:
thanks for the well documented pr @tommie-lie!

tommie-lie:
I just rebased the PR to the latest master and adapted the new CD workflow to use poetry for PyPI publication.

tommie-lie:
Glad to see some cleanup work being done again in mainline. Is there anything I can do to move this PR forward? It has not even been approved to run the tests. Do you need more information?
Should I rebase the PR again? I'd like to have some feedback before putting more time into the PR in vain.

monkut:
pipenv was just introduced into the development flow with the release of 0.55.0, and at the moment pipenv is not causing any issues, and the publishing automation is already in place.

Unless there is another feature or issue that would require this, it's my opinion that this be held off until at earliest the release after next (roughly that would be Q1 of 2023) if there is support from other devs.

PR #1103, opened 2022-01-13
only run publish workflow when creating new version tags

wrboyce:


.github/workflows/cd.yaml
```diff
@@ -1,11 +1,11 @@
 name: CD
 
 on:  # yamllint disable-line rule:truthy
-  create:
-    tags: ["v?[0-9]+.[0-9]+.[0-9]+"]
+  create
 
 jobs:
   publish:
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
```
javulticat, line 8: Should this be `'refs/tags'` instead of `'refs/tags/v'` since our tags currently do not start with `v` (ex: [0.54.1](https://github.com/zappa/Zappa/releases/tag/0.54.1))?

FWIW, the [`pypa/gh-action-pypi-publish` documentation uses this in their examples](https://github.com/pypa/gh-action-pypi-publish#usage):

`if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')`

Should we use that as well?  I'm not sure of the significance of excluding `github.event_name == 'push'`?

javulticat:
Will this create only a `py3` wheel in `pypi` or will it create a ` py2.py3` wheel (or both)?  It looks like for 0.54.1, this automated logic created a `py2.py3` wheel, which seems incorrect, since we do not support Python 2.  AFAIK, we should try to ensure that we are only publishing a `py3` wheel.  Historically, that's all we've published manually.

Examples:

## 0.54.1 (automated - py3 and py2.py3)

image

## 0.54.0 (manual - just py3)

image

monkut:
Handled in https://github.com/zappa/Zappa/issues/1152

cloisng.

PR #1105, opened 2022-02-03
@task should throw an error if def arguments are not json serializable, this will cause code to crash locally if it will crash when deployed to actual lambdas

jottenlips:


jottenlips:
going to close until I write a test

Issue #847, opened 2021-02-20
[Migrated] Add support for REQUEST type Lambda Authorizers


jneves:
Originally from: https://github.com/Miserlou/Zappa/issues/2093 by [tommie-lie](https://github.com/tommie-lie)

## Description
Support for REQUEST authorizers was almost there, as the `type` authorizer property was mostly passed through to the CloudFormation templates. This PR extends the handler middleware to also treat events of type REQUEST as authorization events. It also extends the template creation to treat TOKEN and REQUEST type authorizers equal (defaulting to TOKEN just as before).
The most significant change is in the logic how the authoriz*ation* type for API routes is derived from the authoriz*er* type: Instead of assuming CUSTOM if the type is not specified, the authorizer type correctly defaults to TOKEN. The authorization type is then mapped to CUSTOM for both, TOKEN and REQUEST authorizers.

Added unit tests and tested on live environments on AWS.

## GitHub Issues
Fixes #1159

tsebastiani:
is there any update regarding this PR? Would be super helpful to have this feature merged.

PR #1113, opened 2022-02-25
Add support for authorizer REQUEST type

ppartarr:
## Description
AWS gateway supports the use of Lambda authorizers to perform authorization (grant access to certain resources) based on a given authentication strategy (OpenID connect, basic auth, SAML etc). AWS Gateway offers two kinds of authorizers: REQUEST & TOKEN. 

TOKEN authorizers receive the bearer token, like a JWT or a OAuth token.

REQUEST authorizers receive a combination of headers, query strings, stage variables and context variables.

Zappa only supports TOKEN authorizers, which is quite limiting since some authentication schemes may depend on information provided in the request.

This PR adds support for REQUEST authorizers. The authorizer config now takes a new `type` field as well as an `identity_sources` object which can be a combination of the various request expressions needed for the `REQUEST` authorizer (header, query string, stage variable, or context). We don't explicity support multi stage variable since fields can appear multiple times e.g.
```
"authorizer": {
  "type": "REQUEST",
  "function": "auth.lambda_handler",
  "identity_sources": {
    "headers": ["Authorization", "Host"],
    "stage_variables": ["prod", "dev"]
  },
}
```

For backwards compatibility, if the `type` of an authorizer is not specified, it defaults to `TOKEN`.

## GitHub Issues

https://github.com/zappa/Zappa/issues/847
https://github.com/Miserlou/Zappa/issues/1159

### DIFF
--- a/README.md
+++ b/README.md
@@ -903,19 +903,18 @@ to change Zappa's behavior. Use these at your own risk!
         "iam_authorization": false, // optional, use IAM to require request signing. Default false. Note that enabling this will override the authorizer configuration.
         "include": ["your_special_library_to_load_at_handler_init"], // load special libraries into PYTHONPATH at handler init that certain modules cannot find on path
         "authorizer": {
-            "type": "request", // Authorizer type, request or token (Default 'token')
+            "type": "REQUEST", // Authorizer type, REQUEST or TOKEN (Default 'TOKEN')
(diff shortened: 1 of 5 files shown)


ppartarr:
@hellno @javulticat can this PR get a review?

monkut:
@ppartarr Can you check why this is failing?  You may need need to run `make isort` / `make black`.

PR #1119, opened 2022-04-01
Update permissions in deploy.json

henrikhorluck:
## Description
 
After setting up automated deploys with Zappa, it turned out to be necessary to also give
the `lambda:GetAlias` and `lambda:DeleteFunctionConcurrency` permissions.

Specifically see https://github.com/dotkom/onlineweb4/runs/5793924633?check_suite_focus=true
and https://github.com/dotkom/onlineweb4/runs/5793839836?check_suite_focus=true.


monkut (approved):

souravjamwal77 (approved):
LGTM

Issue #1127, opened 2022-04-25
Handle optional space in x-forwarded-for header


naresh-uplight:
https://github.com/zappa/Zappa/blob/a1e842cf95e5bd99a84e89d5eaaed42ff82f57eb/zappa/wsgi.py#L105

As per the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax) the whitespace is optional and I have a case where the request errors because of that.

monkut:
handled in pr #1127 closing.

monkut:
released in 0.56.0, closing.

PR #1128, opened 2022-04-25
Handle optional whitespace in X-Forwarded-For header

naresh-uplight:
## Description
Whitespace in X-Forwarded-For is optional and therefore should handle the case when there is no whitespace

## GitHub Issues
https://github.com/zappa/Zappa/issues/1127

### DIFF
--- a/zappa/wsgi.py
+++ b/zappa/wsgi.py
@@ -102,6 +102,7 @@ def create_wsgi_request(
     if "," in x_forwarded_for:
         # The last one is the cloudfront proxy ip. The second to last is the real client ip.
         # Everything else is user supplied and untrustworthy.
+        x_forwarded_for = x_forwarded_for.replace(", ", ",")
         remote_addr = x_forwarded_for.split(", ")[-2]
     else:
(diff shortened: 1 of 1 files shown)


zappa/wsgi.py
```diff
@@ -103,7 +103,7 @@ def create_wsgi_request(
         # The last one is the cloudfront proxy ip. The second to last is the real client ip.
         # Everything else is user supplied and untrustworthy.
         x_forwarded_for = x_forwarded_for.replace(", ", ",")
-        remote_addr = x_forwarded_for.split(", ")[-2]
+        remote_addr = x_forwarded_for.split(",")[-2]
```
monkut, line 106: Technically there could still be spaces here... would this be cleaner?

```
forwarded_for_addresses = [addr.strip() for addr in x_forwarded_for.split(",")]
remote_addr = forwarded_for_addresses[-2]
```

monkut:
Handled in PR:
https://github.com/zappa/Zappa/pull/1163

Closing.

Issue #1132, opened 2022-05-10
Not recognizing virtaulenv created with pyenv.


iml1111:
## Context
If .python-version does not exist in the current path, the virtual environment of pyenv is not recognized.

## Expected Behavior

```shell
> zappa update dev
Calling update for stage dev..
Downloading and installing dependencies..
...
```

## Actual Behavior
```shell
# .python-version is not in that path, but pyenv works correctly.
> pwd
/Users/username/Documents/GitHub/myenv/myproject
> pyenv versions
* my-env (set by /Users/username/Documents/GitHub/my-env/.python-version)
...

> zappa update dev
Calling update for stage dev..
Error: Zappa requires an active virtual environment!
Learn more about virtual environments here: http://docs.python-guide.org/en/latest/dev/virtualenvs/
```

## Possible Fix
In addition to detecting the .python-version file in the current path, you can check the current virtual environment by using the `pyenv version` command.

## Steps to Reproduce

1.
2.
3.

## Your Environment

* Zappa version used: 0.54.1
* Operating System and Python version: 3.9.12
* The output of `pip freeze`:
```
argcomplete==2.0.0
awscli==1.22.87
boto3==1.21.32
botocore==1.24.32
certifi==2021.10.8
cfn-flip==1.3.0
charset-normalizer==2.0.12
click==8.1.2
colorama==0.4.3
docutils==0.15.2
durationpy==0.5
Faker==13.3.4
Flask==2.1.1
flask-validation-extended==0.1.7
future==0.18.2
hjson==3.0.2
idna==3.3
importlib-metadata==4.11.3
itsdangerous==2.1.2
Jinja2==3.1.1
jmespath==1.0.0
kappa==0.6.0
MarkupSafe==2.1.1
placebo==0.9.0
pyasn1==0.4.8
python-dateutil==2.8.2
python-dotenv==0.20.0
python-slugify==6.1.1
PyYAML==5.4.1
requests==2.27.1
rsa==4.7.2
s3transfer==0.5.2
six==1.16.0
slack-sdk==3.15.2
text-unidecode==1.3
toml==0.10.2
tqdm==4.64.0
troposphere==4.0.0
urllib3==1.26.9
Werkzeug==2.1.1
wsgi-request-logger==0.4.6
zappa==0.54.1
zipp==3.8.0
```
* Link to your project (optional): I don't think this is necessary.
* Your `zappa_settings.json`:  I don't think this is necessary.

monkut:
[...]

PR #1153, opened 2022-07-18
Add feature pyenv virtualenv detecting .python-version file

iml1111:
## Description
 
The pyenv part was improved in the logic to detect virtualenv.
Previously, the version file (.python-version) of the current path running zappa was found.
But now, it actually detects the version-file recognized by pyenv and executes the command based on that file.

## GitHub Issues

- #1132


monkut:
need to run `make black`, `make isort` to pass the linter.

zappa/core.py
```diff
@@ -511,20 +511,34 @@ def get_current_venv():
...
+        try: # progma: no cover
+            subprocess.check_output(["pyenv", "help"], stderr=subprocess.STDOUT)
+            pyenv_available = True
+        except OSError:
+            pyenv_available = False
+
+        if pyenv_available: # progma: no cover
+            # minor fix find pyenv version-file (.python-version)
+            # Related: https://github.com/zappa/Zappa/issues/1132
+            version_file_path = subprocess.check_output(
```
monkut, line 526: `pyenv version-file` seems to return parent directory `.python-version` files if one doesn't exist in the current directory.

The current implementation appears to only look in the executed directory (?).
The `.python-version` file of the parent may not be the correct version.

Do we need to use the `.python-version` file?  Since we have the executable path, would something like this be better?

(could be cleaner, but just as an example)
```
...
env_name = subprocess.check_output([bin_path, "--version"]).strip().split(" ")[-1]
```

env_name appears to be the python version number, if that's the case maybe we should change the variable name to something like, `python_version`?

monkut: Or even just parse the python version from the bin_path?

iml1111: The existing logic detects venv through pyenv only when a `.python-version` file exists in the current directory.

However, since pyenv can be used even if the `.python-version` file does not exist, the location of the two conditions has been changed. In other words, the `.python-version` directory can be used even if it does not exist in the current path.

In some projects, I've seen multiple repos use a single `.python-version` in the parent directory. However, it will still behave the same if the file is located in the current path as in the previous version.

iml1111: > Or even just parse the python version from the bin_path?
> 
> Is there a case where `pyenv which python` would give you the wrong version?
The existing pyenv detection logic is dependent on the `.python-version` file, so I implemented it with care only in this part. I couldn't find any problems with `pyenv which python`. However, it would be cool if the structure was changed as follows so that additional functions could be added later.


```python

        if pyenv_available:  # progma: no cover
            # minor fix find pyenv version-file (.python-version)
            # Related: https://github.com/zappa/Zappa/issues/1132
            version_file_path = subprocess.check_output(["pyenv", "version-file"]).decode("utf-8").strip()

            if ".python-version" in version_file_path:
                with open(version_file_path, "r") as f:
                    # minor fix in how .python-version is read
                    # Related: https://github.com/Miserlou/Zappa/issues/921
                    env_name = f.readline().strip()
                bin_path = subprocess.check_output(["pyenv", "which", "python"]).decode("utf-8")
                venv = bin_path[: bin_path.rfind(env_name)] + env_name
                return venv
                
            else: # bin_path
            # DO SOMETHING <<

        return None

```

iml1111: > env_name appears to be the python version number, if that's the case maybe we should change the variable name to something like, python_version?

In the .python-version file, the name of venv is often used in addition to the simple python version number. So, while I was thinking here too, I chose `env_name`, but if you want, I will change it to `python_version`.

monkut: Yes, I understand that the existing code uses the `.python-version` file, but reviewing this pr and the pre-existing code, it looks like retrieving the version from the `.python-version` file is unnecessary.

It is only used as a _search string_ for correcting the bin_path:

```venv = bin_path[: bin_path.rfind(env_name)] + env_name```

In this case, it's isn't necessary to use the `.python-version`, and we can use the path discovered by `pyenv which python`, using the right most match.

Could we use something like this to bypass the need to use the `.python-version` file?

```
import re

python_version_match_regex = re.compile("(3\.\d+\.\d+)")

p = "/home/3.8.1/sample/.pyenv/versions/3.8.5/bin/python"
matches = [m for m in python_version_match_regex.finditer(p)]  # get all matches
match_count = len(matches)
last_match = matches[-1]
_, last_index = last_match.span()  # get position of last match
venv = p[:last_index]
```

Comment (iml1111) [reply zappa/core.py]: If I understood correctly, I have one concern.

This is because, in the `.python-version` file, various virtual environment names specified by the user can be specified in addition to the simple python version. For example, the example I am using is as follows:

```
> pwd
/Users/imiml/Documents/GitHub/imfast

> pyenv version-file
/Users/imiml/Documents/GitHub/imfast/.python-version

> cat /Users/imiml/Documents/GitHub/imfast/.python-version
imfast

> pyenv which python
/Users/imiml/.pyenv/versions/imfast/bin/python

```

For this reason, I think that finding it by pattern matching consisting of simple numbers is risky.

As an alternative, it may be possible if it matches and removes like `/bin/python`. However, there may be other exceptions, so this is also a concern.

Comment (monkut) [reply zappa/core.py]: Interesting, I wasn't aware of that.
Yes, we might be able to remove just the last `/bin/python`.  I'm not sure was well if that's stable.

Comment (monkut) [reply zappa/core.py]: From the pyenv documentation:

> Each Python version is installed into its own directory under $(pyenv root)/versions

We could get the value after `versions` and the system separator may be.

Comment (monkut) [reply zappa/core.py]: this may work:
```
$ pyenv version-name
3.9.6
```

Comment (monkut) [reply zappa/core.py]: something like this?
```
pyenv_root = subprocess.check_output(["pyenv", "root"]).decode("utf-8").strip()
pyenv_version = subprocess.check_output(["pyenv", "version-name"]).decode("utf-8").strip()
venv = str(Path(pyenv_root)) / pyenv_version
```

Comment (iml1111) [reply zappa/core.py]: wow, that's a good idea I will implement it soon.

File: zappa/core.py [hunk 2]
```diff
@@ -521,18 +521,12 @@ def get_current_venv():
             pyenv_available = False
 
         if pyenv_available:  # progma: no cover
-            # minor fix find pyenv version-file (.python-version)
+            # minor fix find pyenv version
```
monkut, line 524: Can you add the comment:

```
# Each Python version is installed into its own directory under $(pyenv root)/versions
# https://github.com/pyenv/pyenv#locating-pyenv-provided-python-installations
```

monkut (commented):
Looks nice and clean!

Can you try to add a testcase for this?

monkut (approved):
Great! 
This needs another approval, but I will work to get it in the next release currently targeted for november.

zappa/core.py
monkut: Added thanks!

hellno (approved):
👍🏽👍🏽 looks good too me!

PR #1154, opened 2022-07-27
add feature (#704) Check if args/kwargs are JSON Serializable while running locally

monkut:
## Description

When running an async @task locally, the args and kwargs are passed as they are (even if they're not JSON Serializable).
This PR runs a `validate_json_seralizable()` check on args/kwargs passed to the function when running locally, calling `json.dumps()` on the arguments to allow related exceptions to propagate.

This allow developers to identify asynchronous task related JSON serialization errors prior to deployment.

## GitHub Issues

#704


zappa/utilities.py
```diff
@@ -592,3 +592,11 @@ def merge_headers(event):
...
         multi_headers[h] = ", ".join(multi_headers[h])
     return multi_headers
+
+
+def validate_json_seralizable(function_args: dict) -> None:
+    try:
+        json.dumps(function_args)
+    except Exception:
+        LOG.error("Arguments expected to be JSON seralizable!")
+        raise  # re-raise initial error
```
hellno, line 602: Why not create a custom exception class here?
- easier to understand for humans
- could spare logging the error because the string `arguments expected...` can be use in the exception 

basic example here: https://github.com/zappa/Zappa/blob/fff5ed8ad2ee071896a94133909adf220a8e47d9/zappa/asynchronous.py#L127

`class UnserializableJsonError(Error)` or similar

tests/tests_async.py
```diff
@@ -94,3 +94,10 @@ def test_async_call_with_defaults(self):
             lambda_function_name="MyLambda",
         )
         lambda_async_mock.return_value.send.assert_called_with(get_func_task_path(async_me), ("qux",), {})
+
+    def test_async_call_arg_not_json_seralizable(self):
```
hellno, line 98: seralizable -> ser**i**alizable
minor typo :) just had to double check myself when typing this word as well :)

(tagging the comment on this line, but there are 7 occurences in this PR)

hellno (changes requested):
great work! ran into these errors myself on AWS, will make the dev experience a bit better 🥳
only two minor comments :)

zappa/utilities.py
monkut: Added `zappa.utilities.UnserializableJsonError`

tests/tests_async.py
monkut: Fixed spelling error.

monkut: @hellno updated.

zappa/utilities.py
monkut: @hellno updated.

thommor (approved):

tests/tests_async.py
```diff
@@ -94,3 +95,13 @@ def test_async_call_with_defaults(self):
...
+
+    def test_async_call_arg_not_json_serializable(self):
+        """Exception is raised when calling an async function locally (not on aws)"""
+        async_me = import_and_get_task("tests.test_app.async_me")
+        unserializable_object = datetime.datetime.now()
+        with self.assertRaises(TypeError):
+            async_me(unserializable_object)
+
+        with self.assertRaises(UnserializableJsonError):
+            async_me(unserializable_object)
```
javulticat, line 107:
```suggestion
        with self.assertRaises(UnserializableJsonError):
            async_me(unserializable_object)
```

Am I missing a reason that the latter assertion is not sufficient by itself?

Having both would seem to just be ensuring `UnserializableJsonError` is a subclass of `TypeError`, which doesn't seem too useful.

IMO, our unit tests should assume that Python internals (ie. class inheritance, the fact that `except ExceptionClass:` catches `ExceptionClass` and all of its subclasses, etc.) work they way they are documented (and are thoroughly tested for us already, in the unit tests of `CPython` 🙂).

zappa/utilities.py
```diff
@@ -592,3 +599,10 @@ def merge_headers(event):
     for h in multi_headers.keys():
         multi_headers[h] = ", ".join(multi_headers[h])
     return multi_headers
+
+
+def validate_json_serializable(function_args: dict) -> None:
```
javulticat, line 604: I might suggest we change the signature of this utility function such that it's more generic and reusable elsewhere if needed.  Other Python objects besides dicts can be serialized into valid JSON-encoded strings.

The simplest way to accomplish this would probably be to allow this function accept any amount of objects (in arg and/or kwarg form) and check them all in one go by creating a simple tuple out of them to pass to `json.dumps`.

```python
def validate_json_serializable(*args: Any, **kwargs: Any) -> None:
    try:
         json.dumps((args, kwargs))
     except ...
```

This also removes the burden on the function caller to convert what are potentially multiple objects into a single data structure beforehand, which, for example, would also allow us to trim a line above where it is being used where we are doing exactly that:

```python
function_args = {"args": args, "kwargs": kwargs}
validate_json_serializable(function_args)
```

becomes:

```python
validate_json_serializable(*args, **kwargs)
```

Alternatively, we could just broaden the function definition and leave all else as-is:

```python
def validate_json_serializable(obj: Any) -> None:
    try:
        json.dumps(obj)
    except ...
```

I slightly prefer the former as it doesn't require the user to have to potentially convert multiple objects into a single data structure before using the function (like we are having to do now, by converting args+kwargs to a dict before passing to this function).  WDYT?

tests/tests_async.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf8 -*-
+import datetime
```
javulticat, line 1: Nit: We can probably find an unserializable object without needing to import another library.  For example, `object()` isn't serializable 🙃

javulticat (commented):
Just a few suggestions for very minor things - great work! 🥇

tests/tests_async.py
monkut: Thanks for the review.
Yes, the `TypeError` check is unnecessary.  I've committed the suggestion!

zappa/utilities.py
monkut: My focus was this single use case, but that's a good point.  If we include it in `utilities.py` it should support more wide usage.

Updated!

tests/tests_async.py
monkut: @javulticat no problem, updated!

hellno (approved):
looks good to me!

PR #1162, opened 2022-08-08
Replace prints with logger calls.

jait:
## Description
Replace prints with logger calls. Using print() is bad in frameworks and libraries because then there is no way for the application to control the logging. These prints ended up always in CloudWatch logs, no matter what logging level was used.


thommor (approved):

monkut:
It looks like there are a lot of remaining prints in the codebase, can you look into and handle what you can in this PR?

zappa/handler.py
```diff
@@ -260,9 +260,9 @@ def _process_exception(cls, exception_handler, event, context, exception):
             try:
                 handler_function = cls.import_module_and_get_function(exception_handler)
                 exception_processed = handler_function(exception, event, context)
-            except Exception as cex:
-                logger.error(msg="Failed to process exception via custom handler.")
-                print(cex)
+            except Exception:
+                logger.exception("Failed to process exception via custom handler.")
```
monkut, line 264: output the exception and output the error message.

```
...
except Exception as e:
    logger.exception(e)
    logger.error("Failed to process exception via custom handler.")
```

zappa/handler.py
```diff
@@ -383,8 +383,7 @@ def handler(self, event, context):
             whole_function = event["command"]
             app_function = self.import_module_and_get_function(whole_function)
             result = self.run_function(app_function, event, context)
-            print("Result of %s:" % whole_function)
-            print(result)
+            logger.debug("Result of %s: %s", whole_function, result)
```
monkut, line 386: No need to combine the result and the message, sometimes we want this output separate.

zappa/handler.py
```diff
@@ -474,8 +473,7 @@ def handler(self, event, context):
             app_function = self.import_module_and_get_function(whole_function)
             if app_function:
                 result = self.run_function(app_function, event, context)
-                logger.debug("Result of %s:" % whole_function)
-                logger.debug(result)
+                logger.debug("Result of %s: %s", whole_function, result)
```
monkut, line 476: No need to combine the result and the message, sometimes we want this output separate.

monkut (changes requested):

zappa/handler.py
jait: This is usually unnecessary. logger.exception() logs the exception, too, and gives you a traceback after the error message.

zappa/handler.py
jait: If you image a cloud environment with lots of lambdas running in parallel, it can be difficult to connect isolated `logger.debug(result)` log prints to the correct context. What would be the reason for wanting the result in a separate log message?

zappa/handler.py
jait: If you image a cloud environment with lots of lambdas running in parallel, it can be difficult to connect isolated logger.debug(result) log prints to the correct context. What would be the reason for wanting the result in a separate log message?

If you separate them, you could end up with logs looking e.g. like this:

> Result of a
> Result of a
> Result of b
> True
> Result of a
> True
> False
> Result of c
> True
> False

PR #1177, opened 2022-09-22
Correction to README.

willrogers:
The `exclude` example in the documentation was (still) inaccurate.

See #422 and #1052.


monkut (approved):

souravjamwal77 (approved):
Thanks, exclude previously was confusing to understand while looking from git's ignore point of view.

Issue #1140, opened 2022-07-03
Allow any Python version if using Docker image for Lambda


gaborschulz:
## Context
AWS allows users to work with custom images. These custom images can use any version of Python, not just the predefined ones, like 3.7, 3.8 or 3.9. E.g. I could build my own image using Python 3.10 or 3.11alpha and use that custom image for deployment. However, even though this is a completely valid setup, Zappa does not allow this to happen if I have a virtual environment with Python 3.10 or 3.11.

## Expected Behavior
Allow the usage of any Python version if I am using a custom image for deployment.

## Actual Behavior
The deployment fails because of an unsupported Python version.

## Possible Fix
Skip the environment version check if a custom image is used for deployment

## Steps to Reproduce
1. Create a Python 3.10 environment
2. Use a custom Docker image for deployment with Python 3.10
3. While this is a completely valid setup, Zappa deployment will fail

## Your Environment
* Zappa version used: 0.54.1
* Operating System and Python version: Linux, Python 3.10

monkut:
Zappa supports 3.7 through 3.9.  So you'll need to run zappa from one of those environments, however, I can't seem to find a restriction anywhere on the version used in docker.

Can you share the error?

gaborschulz:
That seems to be the actual issue. I would need to have an environment on Python 3.10 for development and building the container. Then, I would need to switch the environment to a 3.9 just to run `zappa update` and then switch back to the 3.10 environment to continue developing. That becomes very cumbersome, eventually. 

In understand that it would be a relatively complex refactoring as the version check is actually done in the `__init__.py` and performed before anything else is done. 

It would be great to move this to another section of the code and only perform it if the configuration does not specify a Docker image.

monkut:
@gaborschulz I've attempted to add a solution for this case in the PR below.

Let me know if you believe this will resolve your issue.

https://github.com/zappa/Zappa/pull/1180

gaborschulz:
@monkut, that's amazing, I think that's exactly the fix I was looking for. Thank you so much.

There's only one minor thing that would need to be checked:

Pip won't let Zappa be installed inside the Docker container if it's higher than Python 3.9 because of this line in `setup.py`:
```
python_requires=">=3.7, <3.10",
```

Is it possible to change this line to 
```
python_requires=">=3.7",
```

monkut:
@gaborschulz Thanks for the feedback, updated!

monkut:
[...]

PR #1180, opened 2022-10-06
Bypass python version runtime check when code run in docker

monkut:
## Description
Currently zappa performs a supported python version check on import.
In the case where zappa is used with docker, zappa will not allow unsupported python versions to be used.

This PR by-passes the full SUPPORTED_VERSION check on import when it detects that it is being run within docker and *ONLY* checks the `MINIMUM_SUPPORTED_MINOR_VERSION` (newly defined/added) when docker is detected.

This uses the `/proc/1/cgroup` check mentioned here to determine if code is being run in docker:
https://stackoverflow.com/a/20012536/24718

## GitHub Issues
https://github.com/zappa/Zappa/issues/1140

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,4 +73,7 @@ test-async:
 test-general:
 	nosetests tests/tests.py --with-coverage --cover-package=zappa --with-timer
 
-tests: clean test-docs test-handler test-middleware test-placebo test-async test-general
+test-utilities:
+	nosetests tests/tests_utilities.py --with-coverage --cover-package=zappa --with-timer
+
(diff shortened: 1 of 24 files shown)


souravjamwal77:
@monkut I think we need to merge master into this to keep it's code updated.

monkut:
@souravjamwal77 There doesn't appear to be an issue in github. 
If you have time can you review the code?

souravjamwal77 (approved):
I couldn't test this inside the docker build because the aws-lambda image on the docker hub has no python3.10 version yet. If they publish a new version then I will test it out again. Approving for now.

goya813 (approved):
Looks Good!

PR #1184, opened 2022-10-27
fix(image): use lambda handler override image command

kentio:


monkut:
Is there a related issue with this? 

If your adding a new cli option the README will need to be updated.

kentio:
> Is there a related issue with this?
> 
> If your adding a new cli option the README will need to be updated.

not yet, when I have time I will add

Issue #1191, opened 2022-11-15
Feature Request: Package boto/botocore optionally


chris-erickson:
## Context
From the docs, Zappa excludes boto from the final package because lambda has it already. Unfortunately the AWS version is ancient and tools like `aioboto` rely on functionality of newer versions. It's unclear how one can easily get boto packaged along with app files. Is there a way forward to get boto packaged with deploys, if optionally?

## Expected Behavior
Let me package whatever version of boto I want.

## Actual Behavior
It's excluded [per the docs](https://github.com/zappa/Zappa#how-zappa-makes-packages).

## Possible Fix
Give a config option to skip this, default behavior is to keep current functionality.

## Steps to Reproduce
Attempt to deploy an app with a modern version of aioboto like `aioboto3==10.1.0`. Lambda will fail to start when the aioboto import is reached and it tries to access a method on boto/botocore that doesn't exist on recent versions.

## Your Environment
* Zappa version used: Latest
* Operating System and Python version: Lambda, 3.9

monkut:
Here are the current default excludes:
```
["boto3", "dateutil", "botocore", "s3transfer", "concurrent"]
```
https://github.com/zappa/Zappa/blob/master/zappa/cli.py#L2443

It's unintuitive, but it appears you should be able to optionally include all these by settings `exclude=[]` so that the defaults are not used.

monkut:
an update to exclude handling will be handled in the issue below:
https://github.com/zappa/Zappa/issues/1241

closing this issue.

PR #1192, opened 2022-11-16
Clarify how `exclude` works by default

chris-erickson:
## Description
Clarifies documentation about how `exclude` works, as it relates to Boto.

## GitHub Issues
https://github.com/zappa/Zappa/issues/1191
https://github.com/zappa/Zappa/issues/643


README.md
```diff
@@ -361,7 +361,7 @@ In addition, Zappa will also automatically set the correct execution permissions
...
 
 * Set `slim_handler` to `True` to upload a small handler to Lambda and the rest of the package to S3. For more details, see the [merged pull request](https://github.com/Miserlou/Zappa/pull/548) and the [discussion in the original issue](https://github.com/Miserlou/Zappa/issues/510). See also: [Large Projects](#large-projects).
-* Use the `exclude` or `exclude_glob` setting and provide a list of patterns to exclude from the archive. By default, Zappa will exclude Boto, because [it's already available in the Lambda execution environment](http://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html).
+* Use the `exclude` or `exclude_glob` setting and provide a list of patterns to exclude from the archive. By default, Zappa will exclude Boto, because [it's already available in the Lambda execution environment](http://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html). This will cause your application to fall back to the Boto tools that ship with AWS Lambda, which can be quite old. If this is a problem for your deployments, override `exclude` defaults with your own, removing just those: `"exclude": ["dateutil", "s3transfer", "concurrent"]`.
```
monkut, line 364: * Use the `exclude` or `exclude_glob` setting and provide a list of patterns to exclude from the archive. By default, 
when `slim_handler=False` *exclude* defaults to `["boto3", "dateutil", "botocore", "s3transfer", "concurrent"]`, because [it's already available in the Lambda execution environment](http://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html). This will cause your application to fall back to the versions that are shipped with AWS Lambda. If this is a problem, override `exclude` defaults by supplying a value.  (Example: `"exclude": ["dateutil", "s3transfer", "concurrent"]`).

How's this, some minor clarification/cleanup.

monkut (changes requested):

README.md
monkut: `quite old`/`very old` statements are ambiguous.  I feel that the `... will cause your application to fall back to the versions that are shipped with AWS Lambda.` statement is clear enough, but if you want to be clearer, perhaps link to the page that lists the versions used.

Maybe:

* Use the `exclude` or `exclude_glob` setting and provide a list of patterns to exclude from the archive. By default, 
when `slim_handler=False` *exclude* defaults to `["boto3", "dateutil", "botocore", "s3transfer", "concurrent"]`, because [it's already available in the Lambda execution environment](http://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html). This will cause your application to fall back to the [versions included with the selected AWS Lambda runtime](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html). If this is a problem, override `exclude` defaults by supplying a value.  (Example: `"exclude": ["dateutil", "s3transfer", "concurrent"]`).

chris-erickson: Love it!

Issue #1190, opened 2022-11-14
Incorrect hosted zone match


sridhar562345:
https://github.com/zappa/Zappa/blob/9b2a98ccaebe271e614c0af1bcf301ea6facb765/zappa/core.py#L3197-L3208

Suppose there are two hosted zones `api.in` and `beta.api.in` and domain name specified is `some-beta.api.in`

The current logic selects hosted `beta.api.in` as the hosted zone for the specified domain.

monkut:
What is the expected result?

sridhar562345:
> What is the expected result?

It should select `api.in` as the best match zone

monkut:
@sridhar562345 I attempted an improvement PR, can you check it out and let me know if it meets your expectations?

sridhar562345:
@monkut

I was actually working on it as well and I guess I have a better logic to verify this.

Here is my logic
```py
@staticmethod
def get_best_match_zone(all_zones, domain):
    """Return zone id which name is closer matched with domain name."""

    # Related: https://github.com/Miserlou/Zappa/issues/459
    public_zones = [zone for zone in all_zones['HostedZones'] if not zone['Config']['PrivateZone']]

    zones = {zone['Name'][:-1]: zone['Id'] for zone in public_zones if zone['Name'][:-1] == domain}

    if not zones:
        zones = {zone['Name'][:-1]: zone['Id'] for zone in public_zones if '.' + zone['Name'][:-1] in domain}

    if zones:
        keys = max(zones.keys(), key=lambda a: len(a))  # get longest key -- best match.
        return zones[keys]
    else:
        return None
```

monkut:
Thanks!

monkut:
released in 0.56.0, closing.

PR #1193, opened 2022-11-17
Improve `get_best_match_zone()` to match by most matched components instead of by length of domain

monkut:
## Description
Current `get_best_match_zone()` chooses the best match by length, so the expectation will result in error where a match is a partial match.

When zones are:

- `example.com`
- `beta.example.com`
 
And `some-beta.example.com` is given as the *domain*, the expected match is `example.com`, but `beta.example.com` will match.

This PR picks the zone with the most matches instead of length

## GitHub Issues
https://github.com/zappa/Zappa/issues/1190

### DIFF
--- a/README.md
+++ b/README.md
@@ -928,7 +928,7 @@ to change Zappa's behavior. Use these at your own risk!
         ],
         "endpoint_configuration": ["EDGE", "REGIONAL", "PRIVATE"],  // Specify APIGateway endpoint None (default) or list `EDGE`, `REGION`, `PRIVATE`
         "exception_handler": "your_module.report_exception", // function that will be invoked in case Zappa sees an unhandled exception raised from your code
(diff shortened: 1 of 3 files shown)


souravjamwal77 (approved):
Looks good.

goya813:
Looks good p-r!!

I wanted the following test cases.
```
zone = Zappa.get_best_match_zone(
    all_zones={
        "HostedZones": [
            {
                "Name": "beta.example.com.",
                "Id": "zone-incorrect",
                "Config": {"PrivateZone": False},
            },
            {
                "Name": "example.com.",
                "Id": "zone-correct",
                "Config": {"PrivateZone": False},
            }
        ]
    },
    domain="example.com",
)
```

monkut:
@goya813 

Thanks for the review, nice catch!

I've added a testcase, and updated the code to account for that case.

goya813 (approved):
Looks Good!

PR #1194, opened 2022-11-21
Added posix path for tarball(Issue #516)

muddi900:
Fixing issue #516

It outputs a better posix.

There is a test failing but the error is a connection Timeout error that seems to be nothing to do with this change

```
botocore.exceptions.ConnectTimeoutError: Connect timeout on endpoint URL: "http://169.254.169.254/latest/meta-data/iam/security-credentials/"
```


monkut:
We use isort/black for formatting, you don't need to change the format manually.

monkut:
@muddi900 Can you add a testcase for this?

PR #1198, opened 2022-11-29
fix(lambda config): fix lambda is {}

kentio:
## Description
 
labmda cofnig is `{}`,  and will not be updated

## GitHub Issues

no issue

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -2183,11 +2183,13 @@ class Zappa:
                 description_kwargs[key] = value
             elif key == "LambdaConfig":
                 for lckey, lcvalue in value.items():
+                    # update needed config
                     if lckey in LambdaConfig:
                         value[lckey] = LambdaConfig[lckey]
                 print("value", value)
                 description_kwargs[key] = value
-        if "LambdaConfig" not in description_kwargs:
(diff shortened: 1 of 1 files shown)


sridhar562345:
@kentio 
I don't see any need of changing that line.

zappa/core.py
```diff
@@ -2183,11 +2183,13 @@ def update_cognito(self, lambda_name, user_pool, lambda_configs, lambda_arn):
                 description_kwargs[key] = value
             elif key == "LambdaConfig":
                 for lckey, lcvalue in value.items():
+                    # update needed config
                     if lckey in LambdaConfig:
                         value[lckey] = LambdaConfig[lckey]
                 print("value", value)
                 description_kwargs[key] = value
```
kentio, line 2190: @sridhar562345 When associated with User Pool, Lambda Config is {}, so it will never add a user-defined configuration.

monkut:
I don't see what the issue is.
Can you add a testcase to reproduce the problem?

hellno:
Just going through open PRs and doing some housekeeping. 
@kentio if you want to get this merged, happy to chat here or in Zappa Slack to get it done.

Issue #1201, opened 2022-12-01
FileNotFoundError in running_in_docker


hassa:
## Context

In `running_in_docker` if there is no `/proc/1/cgroup`, like on OSX for example, a FileNotFoundError is raised and zappa can't be imported.

We do some stuff with SQS and zappa's asynchronous tasks so we import zappa even during dev.

## Expected Behavior

FileNotFoundError should be caught and running_in_docker should return False.

## Actual Behavior

A FileNotFoundError is raised.

## Possible Fix

FileNotFoundError should be caught and running_in_docker should return False.

## Steps to Reproduce

1. Be on OSX
2. import zappa
3.

## Your Environment

* Zappa version used: 0.56.0
* Operating System and Python version: OSX 12.6, Python 3.9 and 3.11.0
* The output of `pip freeze`:
* Link to your project (optional):
* Your `zappa_settings.json`:

rogerhoward:
Having the same issue, but triggered by "zappa init" even; eg:

1. create a virtualenv and activate it
2. install zappa ("pip install zappa")
3. run "zappa init"

monkut:
Thanks for reporting this, we'll get a fix out soon!

pkaufhol:
Same issue on osx!

monkut:
Method was changed, and fixed in the 0.56.1 release, please check.

PR #1203, opened 2022-12-01
fix FileNotFoundError in running_in_docker (#1201)

monkut:
running_in_docker() check introduced in 0.56.0, and assumes that `/proc/1/cgroup` is always on the system (linux system assumed), this PR updates to check for the expected file existence before attempting to read it's content.

- :bug: fix running_in_docker()
- :arrow_up: increments version to 0.56.1
- :pencil: updates CHANGELOG.md with 0.56.1 and with #1201

Fixes:
- FileNotFoundError in running_in_docker (#1201)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Zappa Changelog
 
+## 0.56.1
+
+* FileNotFoundError in running_in_docker (#1201)
+
 ## 0.56.0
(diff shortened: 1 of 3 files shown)


monkut:
Closing this in place of cleaner approach defined in this PR:
https://github.com/zappa/Zappa/pull/1204

PR #1204, opened 2022-12-02
Alternative way to check if running in Docker

gaborschulz:
## Description
 
The `running_in_docker` function can lead to unpredictable results, in certain cases it does not recognize that it's running inside docker and leads to the version check being in place. This produces errors in build pipelines.

## GitHub Issues

https://github.com/zappa/Zappa/issues/1140
https://github.com/zappa/Zappa/issues/1201


zappa/__init__.py
```diff
@@ -1,35 +1,27 @@
...
     """
     Determine if zappa is running in docker.
     - When docker is used allow usage of any python version
     """
-    # https://stackoverflow.com/a/20012536/24718
-    cgroup_content = Path("/proc/1/cgroup").read_text()
-    in_docker = "/docker/" in cgroup_content or "/lxc/" in cgroup_content
-    return in_docker
+    # https://stackoverflow.com/questions/63116419
+    running_in_docker_flag = os.getenv("ZAPPA_RUNNING_IN_DOCKER", "False").lower() in ("true", "1", "t")
```
monkut, line 13: distutils has a strtobool function, it is deprecitated, but has slightly more expanded options.
https://github.com/python/cpython/blob/3.10/Lib/distutils/util.py#L308

```
in_docker = os.getenv("ZAPPA_RUNNING_IN_DOCKER", "False").lower() in ('y', 'yes', 't', 'true', 'on', '1')
```

zappa/__init__.py
```diff
@@ -1,35 +1,27 @@
...
-        f"Zappa (and AWS Lambda) support the following versions of Python: {formatted_supported_versions}"
-    )
-    raise RuntimeError(err_msg)
-elif running_in_docker() and sys.version_info.minor < MINIMUM_SUPPORTED_MINOR_VERSION:
-    # when running in docker enforce minimum version only
-    err_msg = (
-        f"This version of Python ({sys.version_info.major}.{sys.version_info.minor}) is not supported!\n"
-        f"Zappa requires a minimum version of 3.{MINIMUM_SUPPORTED_MINOR_VERSION}"
-    )
+    err_msg = "This version of Python ({}.{}) is not supported!\n".format(
```
monkut, line 22: Since zappa doesn't support <=3.6 let's keep the MINIMUM_SUPPORTED_MINOR_VERSION check.

monkut (changes requested):

monkut:
If the current check is unstable/unreliable, using a environment variable is much cleaner!  
Thanks!

zappa/__init__.py
gaborschulz: Good point, thanks for this

zappa/__init__.py
gaborschulz: You're right, I've added this back in

zappa/__init__.py
```diff
@@ -10,11 +11,12 @@ def running_in_docker() -> bool:
     - When docker is used allow usage of any python version
     """
     # https://stackoverflow.com/questions/63116419
-    running_in_docker_flag = os.getenv("ZAPPA_RUNNING_IN_DOCKER", "False").lower() in ("true", "1", "t")
+    running_in_docker_flag = strtobool(os.getenv("ZAPPA_RUNNING_IN_DOCKER", "False").lower())
```
monkut, line 14: Sorry, my comment wasn't clear. 

My suggestion was to use the conditions used in `strtobool()`, expanding the conditions only, not use `strtobool()`.
Can you update this to:

```
running_in_docker_flag = os.getenv("ZAPPA_RUNNING_IN_DOCKER", "False").lower() in ('y', 'yes', 't', 'true', 'on', '1')
```

monkut (changes requested):

monkut:
Can you pull in the CHANGLOG and version updates in this PR?
https://github.com/zappa/Zappa/pull/1203

Since this is an open issue, I want to get this fix released asap.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Zappa Changelog
 
+## 0.56.1
+
+* Use ZAPPA_RUNNING_IN_DOCKER environment variable to use any Python version inside Docker container (#1140)
```
monkut, line 5: Add the following as well, as this resolves the issue there.

https://github.com/zappa/Zappa/issues/1201

monkut (changes requested):

monkut:
@gaborschulz can you check the lint results and run `make black`, `make isort` to fix formatting issues?

gaborschulz (commented):
Done

zappa/__init__.py
```diff
@@ -1,16 +1,16 @@
+import os
 import sys
-from pathlib import Path
 
 
+# Alternative solution to: https://github.com/zappa/Zappa/issues/1140
```
monkut, line 5: This comment is unnecessary, can you remove it?

monkut (changes requested):
one minor change request

zappa/__init__.py
gaborschulz: Done

monkut (approved):

goya813 (approved):
LGTM!
The test now passes in my Mac Book environment!

Issue #1036, opened 2021-09-14
zappa schedule doesnt work when event function string too long


opqpop:
## Context
Running below zappa_settings.json won't work for Cloudwatch event -> lambda, because the function name is too long and lambda tries to call "aws.lambda_service.pipeline.core.lambda_update_price_col.ma" instead of "aws.lambda_service.pipeline.core.lambda_update_price_col.main", since zappa handler looks at the resource name of the cloud watch event, which has a max char limit

```
    "update_price_col_prod": {
        "aws_region": "us-east-1",
        "profile_name": "default",
        "project_name": "test",
        "events": [{
            "function": "aws.lambda_service.pipeline.core.lambda_update_price_col.main",
            "expression": "cron(0/1 3 * * ? *)"
        }],
        "lambda_description": "update_price_col",
    }
```

Offending code is here: https://github.com/zappa/Zappa/blob/89e99174ded0abdcbc5740c57c25baf93e55d468/zappa/handler.py#L385

## Expected Behavior
Should work regardless of how long the name. We can't refactor our codebase as there's too many files, so we have to rename our files to be shorter if we want to use zappa schedule. For now, most of the devs just stopped using zappa and manually create the lambda / cloudwatch event which is not ideal.

## Actual Behavior

```


[...]

PR #1205, opened 2022-12-03
Added throw error when event function name is long

goya813:
## Description
 

- Added exception handling for cases where event function names are long.
- The problem is that if the function name is too long, the function name is cut off and lambda cannot be called from the Event Bridge.

## GitHub Issues

- https://github.com/zappa/Zappa/issues/1036


zappa/core.py
```diff
@@ -2920,6 +2920,12 @@ def get_scheduled_event_name(self, event, function, lambda_name, index=0):
         Returns an AWS-valid CloudWatch rule name using a digest of the event name, lambda name, and function.
         This allows support for rule names that may be longer than the 64 char limit.
         """
+        function_regex = re.compile("^[._A-Za-z0-9]+$")
```
monkut, line 2923: Can you add a link in the docstring to the rule name requirement?
https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html

image

The expression is shown as:

> [\.\-_A-Za-z0-9]+

Or is this just for the function name?

You can also add min/max with {1, 64} here instead of using "+".

> ^[\.\-_A-Za-z0-9]{1,64}$

https://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html
image

zappa/core.py
```diff
@@ -2930,6 +2936,15 @@ def get_scheduled_event_name(self, event, function, lambda_name, index=0):
             # prefix all entries bar the first with the index
             # Related: https://github.com/Miserlou/Zappa/pull/1051
             name = "{}-{}".format(index, name)
```
monkut, line 2938: What does this `index` and `name` resolve to?

I thought it was function_name and stage, but the variable names here seem to suggest something else.

zappa/core.py
```diff
@@ -2930,6 +2936,15 @@ def get_scheduled_event_name(self, event, function, lambda_name, index=0):
...
+
+        # https://github.com/zappa/Zappa/issues/1036
+        # Error because the name cannot be obtained if the function name is longer than 64 characters
+        if len(name) >= 64:
+            raise EnvironmentError(
+                "Length Constraints:"
+                "Maximum length of 63 ({{expression index}}-{{event['name']}}?-{{event['function']}}):"
```
monkut, line 2945: Ok, this is hard for the user to understand maybe.  

It may be a good idea to include the components used in generating the rule name in the exception, so the user can determine what they need to change.

zappa/core.py
goya813: My understanding is that the process is to avoid duplicate names if there are multiple expressions.
https://github.com/zappa/Zappa/issues/1036

example:
```
{
...
"events": [
            {
                "function": "application.utils.cron.some_func",
                "expressions": [
                    "cron(*/5 22-23 ? * SUN-THU *)",
                    "cron(*/5 0-8 ? * MON-FRI *)"
                ]
            }
        ]
...
}
```

zappa/core.py
goya813: > Can you add a link in the docstring to the rule name requirement?
OK!

zappa/core.py
monkut: This may be an existing issue, but does the _index_ take intoaccount the `STAGE`?  

It appears for this code that the resulting rule name will be the same across all defined stages.

zappa/core.py
```diff
@@ -2919,7 +2919,15 @@ def get_scheduled_event_name(self, event, function, lambda_name, index=0):
         """
         Returns an AWS-valid CloudWatch rule name using a digest of the event name, lambda name, and function.
         This allows support for rule names that may be longer than the 64 char limit.
+
+        function pattern: '^[._A-Za-z0-9]{,63}$'
         """
+        function_regex = re.compile("^[._A-Za-z0-9]{,63}$")
```
monkut, line 2925: You should need the "\." here to allow a "." (period), as "." alone matches all characters except new lines.

monkut: Also, I'm unclear on what `{,63}` does.  I assume that it allows any number of characters up to 63, is that correct?

If that's the case it may be clearer to write `{0,63}`.

goya813: > You should need the "." here to allow a "." (period), as "." alone matches all characters except new lines.

I got the following warning message and removed the escape
> Redundant character escape '\.' in RegExp 

> If that's the case it may be clearer to write {0,63}.
ok!

zappa/core.py
goya813: > It appears from this code that the resulting rule name will be the same across all defined stages.

I understand it the same way.

zappa/core.py
```diff
@@ -2919,7 +2919,17 @@ def get_scheduled_event_name(self, event, function, lambda_name, index=0):
         """
         Returns an AWS-valid CloudWatch rule name using a digest of the event name, lambda name, and function.
         This allows support for rule names that may be longer than the 64 char limit.
+
+        function pattern: '^[._A-Za-z0-9]{,63}$'
```
monkut, line 2923: Can we update this to `{0,63}` for clarity?

goya813: Thank you! fixed

zappa/core.py
```diff
@@ -2919,7 +2919,17 @@ def get_scheduled_event_name(self, event, function, lambda_name, index=0):
         """
         Returns an AWS-valid CloudWatch rule name using a digest of the event name, lambda name, and function.
         This allows support for rule names that may be longer than the 64 char limit.
+
+        function pattern: '^[._A-Za-z0-9]{0,63}$'
```
monkut, line 2923: Shouldn't this be?  

> ^[\._A-Za-z0-9]{0,63}$

zappa/core.py
```diff
@@ -2930,11 +2940,20 @@ def get_scheduled_event_name(self, event, function, lambda_name, index=0):
             # prefix all entries bar the first with the index
             # Related: https://github.com/Miserlou/Zappa/pull/1051
             name = "{}-{}".format(index, name)
+
+        # https://github.com/zappa/Zappa/issues/1036
+        # Error because the name cannot be obtained if the function name is longer than 64 characters
+        if len(name) > max_name_length:
+            raise EnvironmentError(
+                "Too long schedule function name. Please shortening the schedule function name:"
```
monkut, line 2948: `function` is checked above, and will raise an error already.

Do we need to check both the `function`, and the `event['name']`?
It seems that we could just check the resulting `name` only.

goya813: "-" cannot be used for "function", but can be used for "name" Therefore, we have separated these two.

zappa/core.py
monkut: Ok, when using `[]` it's not needed to escape.  Understood.

zappa/core.py
monkut: Can you update the error message for clarity?

> Generated scheduled event name is too long, shorten the function name!

monkut: And include the resulting `name` in the error message.

README.md
```diff
@@ -936,11 +936,11 @@ to change Zappa's behavior. Use these at your own risk!
         "environment_variables": {"your_key": "your_value"}, // A dictionary of environment variables that will be available to your deployed app. See also "remote_env" and "aws_environment_variables". Default {}.
         "events": [
             {   // Recurring events
-                "function": "your_module.your_recurring_function", // The function to execute
+                "function": "your_module.your_recurring_function", // The function to execute (Pattern: [\._A-Za-z0-9]+).
```
monkut, line 939: If `\.` isn't used in the pattern in the code we can remove it from here.

README.md
```diff
@@ -936,11 +936,11 @@ to change Zappa's behavior. Use these at your own risk!
...
                 "expression": "rate(1 minute)" // When to execute it (in cron or rate format)
             },
             {   // AWS Reactive events
-                "function": "your_module.your_reactive_function", // The function to execute
+                "function": "your_module.your_reactive_function", // The function to execute (Pattern: [\._A-Za-z0-9]+).
```
monkut, line 943: If `\.` isn't used in the pattern in the code we can remove it from here.

monkut (approved):

hellno (approved):

javulticat:
Hey there folks - I was looking over this PR this evening and thinking about the issue it aims to solve and I got to wondering... is this PR actually necessary?  What problem is it actually solving?  And for whom?

I know AWS has a character limit for event function names, but should our solution to that be to throw another Exception at our users for each of their event function names that exceed that limit?  Because I believe we resolved the handling of that several years ago in a more elegant way with #1080, and I would've expected that #1080 would've been able to solve the issue referenced here (#1036) as well, since it allowed for event function names of any length to be given to Zappa's config, and it would automatically handle making sure that they ended up with nice, uniquely identifiable names that conformed to AWS limits.

Now, since #1080 was actually merged several months _after_ #1036 was opened, the user who had the complaint in that issue may not have gotten a chance to try the new functionality out that may have solved the issue they were having.  In any case, the user has never once returned to follow up on the issue since they made it over 2 years ago (and if we were on top of our issue pruning, that issue would have been closed as stale over a year ago 😄), so it doesn't seem like it caused them too significant of a problem - perhaps they tried a release with #1080 present in it shortly afterward and that solved their problems after all.

In any case, I'm especially curious then why we felt the need to prioritize handling a 2+ year old stale issue that, in all likelihood, sounds like it may have already been fixed years ago by #1080.  And, that aside, I'm also a little confused by the approach found in this PR (#1205).  If I understand correctly, this PR seems like it would change the behavior introduced in #1080 so that now, a user including a long event function name in a Zappa config will halt the deploy/update process in its tracks, and put the onus on the user to manually rename any/all of their event function names which are too long by AWS standards.  And better still, it looks like the Exception it gives is for the first event function name it finds that is too long, even if there are hundreds.  So, are users just expected to try to run `zappa deploy` hundreds of times and rename the event function names called out by the Exception one-by-one until they no longer hit this new Exception?

I've worked at places where we had literally thousands of event functions with names in our Zappa config that exceeded the AWS character limit, which was one of my motivators for developing #1080 - it was literally faster to create that PR that to rename all of our event functions and find and rename them in all of the places they may have been named in other repos/documentation/wikis/trainings/etc. within my organization.  If we got an error telling us we needed to change a different event function name each time we tried to deploy 5000+ times in a row, that wouldn't exactly be a great user experience (and a great waste of a week).

So, anyway, to conclude my old man shaking fist at sky rambling, my first impression of this PR is that it might be a regression that puts work we've solved years ago for our users right back onto our users' plates. So, what is the compelling reason for why we need to now halt users' deploy processes to explicitly tell them that they need to manually shorten their event function names one-by-one rather than have Zappa continue to shorten all of their event function names for them automatically like it has been doing without incident for years?

Maybe I'm missing something obvious (wouldn't be the first time 😄), but I'm just not sure I understand the thought process that led to this PR being considered necessary to create, for numerous reasons.

Issue #1207, opened 2022-12-06
All APIs not being fetched


DilLip-Chowdary-Codes:
## Context
As mentioned in the [AWS Docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html), we can have default limits of 600, 120 regional, and edge APIs respectively.

But we're fetching only limited (500) APIs, refer to the LOCs below.

https://github.com/zappa/Zappa/blob/73ee393dd48500d321d65e92aba31a5fd83e93ab/zappa/core.py#L2425

https://github.com/zappa/Zappa/blob/73ee393dd48500d321d65e92aba31a5fd83e93ab/zappa/core.py#L2101

## Expected Behavior

To fetch all the available APIs

## Actual Behavior

Fetching only 500 APIs

## Possible Fix

Please update it to fetch all the records.

## Steps to Reproduce

1.
2.
3.

## Your Environment

* Zappa version used:
* Operating System and Python version:
* The output of `pip freeze`:
* Link to your project (optional):
* Your `zappa_settings.json`:

javulticat:
I think you are getting mixed up between AWS's regional account limits and their API response size limits.  The docs you linked do indeed say that an AWS _account_ can have up to 600 regional APIs set up via API Gateway, per AWS region.

However, when calling the `._get_rest_apis()` method from the `boto3` library like we are doing in the places you referenced in the Zappa codebase, [the `boto3` documentation clearly states that a maximum limit of 500 APIs can be returned per each method call of `._get_rest_apis()`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway/client/get_rest_apis.html#get-rest-apis).  So, regardless of how many APIs AWS allows its users to have on their accounts, they still limit their client API to only be able to fetch data on up to 500 AWS APIs per request.

Thus, I'm going to close this, as there is nothing worthwhile we can do about hard limits AWS has set in their client API.

Though also... sorry, but now I'm just really, really curious about this: are you actually deploying more than 500 APIs in a single AWS region using API Gateway (bonus points if you're deploying all 500+ APIs using Zappa) for some... very unique reason?  And if so, what could such a reason for having so many individual APIs possibly be? 🤔

PR #1208, opened 2022-12-06
Fetch all the APIs available instead of limiting to certain num of APIs

DilLip-Chowdary-Codes:
## Description
 
Fetch all the APIs instead of limiting it to 500, related  to issue #1207 
## GitHub Issues

#1207


zappa/core.py
```diff
@@ -2098,9 +2098,21 @@ def get_rest_apis(self, project_name):
         """
         Generator that allows to iterate per every available apis.
         """
-        all_apis = self.apigateway_client.get_rest_apis(limit=500)
+        all_apis = []
 
-        for api in all_apis["items"]:
+        pagination_position = "init"
```
monkut, line 2103: Just thinking since it's yielding the result there's no need to collect all results into a list then yield.
How about yielding results on each response?

Something like:
```
response = self.apigateway_client.get_rest_apis(limit=500)
yield from (api for api in response["items"] if api["name"] == project_name)
while response.get("position"):
    response = self.apigateway_client.get_rest_apis(limit=500, position=response["position"])
    yield from (api for api in response["items"] if api["name"] == project_name)
```

zappa/core.py
```diff
@@ -2422,11 +2434,23 @@ def get_api_id(self, lambda_name):
         except Exception:  # pragma: no cover
             try:
                 # Try the old method (project was probably made on an older, non CF version)
-                response = self.apigateway_client.get_rest_apis(limit=500)
+                rest_apis_list = []
```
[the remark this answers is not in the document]
sridhar562345: @souravjamwal77
Boto3 has a max limit of 500 per get_rest_apis call. I think it's better to generalize the solution instead of hardcoding the value, so we don't have to change it in case AWS increases these limits in the future.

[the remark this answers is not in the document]
monkut: Yes, according to the boto3 docs the `get_rest_apis()` max is 500 and default is 25.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/apigateway.html#APIGateway.Client.get_rest_apis

I don't think it's an issue if we set to the max, 500, do you?

zappa/core.py
monkut: @DilLip-Chowdary-Codes Do you have time to review/update or is it better for one of the core developers to take this over?

sridhar562345 (commented):
seems fine to me

zappa/core.py
```diff
@@ -2098,12 +2098,17 @@ def get_rest_apis(self, project_name):
         """
         Generator that allows to iterate per every available apis.
         """
-        all_apis = self.apigateway_client.get_rest_apis(limit=500)
+        pagination_position = "init"
```
monkut, line 2101: No big issues, however, in terms of code clarity I'm a little adverse to re-purposing an variable.

There are two conditions being evaluated here, if it's the initial and then if there are additional pages.
It's my preference that these two conditions are managed in separate variables, perhaps something like:
```
next_position = None
is_initial = False
while next_position or is_initial:
    response = self.apigateway_client.get_rest_apis(limit=500, position=next_position)
    yield from (api for api in response.get("items", []) if api["name"] == project_name)

    is_initial = False
    next_position = response.get("position", None)
   
    
```

zappa/core.py
```diff
@@ -2098,12 +2098,17 @@ def get_rest_apis(self, project_name):
...
-            if api["name"] != project_name:
-                continue
-            yield api
+        while pagination_position:
+            if pagination_position != "init":
+                response = self.apigateway_client.get_rest_apis(limit=500, position=pagination_position)
+            else:
+                response = self.apigateway_client.get_rest_apis(limit=500)
+
+            yield from [api for api in response.get("items", []) if api["name"] == project_name]
```
monkut, line 2109: This can be be a generator expression in place of a list-comprehension. (generator expression will evaluate and return one by one as opposed to instantiating a list then yielding objects from that list.
```
yield from (api for api in response.get("items", []) if api["name"] == project_name)
```

monkut (commented):
minor style and general comments

sridhar562345 (commented):
review done, now way more meaningful as @monkut suggested

monkut (changes requested):
Ok, looks pretty good.

Can you add a testcase for this, to confirm that we get all results  now (over 500)?

monkut (commented):
Maybe there is something wrong with github diff, but It appears that this PR no longer contains the issue's target fix.

monkut:
Branch is out of sync with the content described in this PR, closing.

Issue #1215, opened 2023-02-06
Error: CloudFormation stack missing, re-deploy to enable updates


kyaryunha:
## Context

My local aws config region is `ap-northeast-2 (Seoul)`, and the gitlab runner's aws region is also `ap-northeast-2 (Seoul)`.
(I can't change the aws region of gitlab runner.)

And the aws region of zappa_settings.json is `us-east-1 (Northern Virginia)`.

In this situation, my local `zappa update stage` works normally, but

When I did `zappa update stage` in Gitlab, I got the following error.

```shell
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the DescribeStackResource operation: Stack 'stack-name-stage' does not exist
CloudFormation stack missing, re-deploy to enable updates
Oh no! An error occurred! :(
```

I have no idea why the above error pops up in the gitlab runner.

Also, when I checked Cloud Formation, Cloud Formation was created well with the corresponding stack name.

I think ubuntu looks for the stack name of cloud formation as the region of aws config, not the aws_region of zappa_settings.json.

## Expected Behavior

Local and gitlab runner `zappa update stage` should match the behavior.

## Actual Behavior
But it didn't work like that only in gitlab runner.

## Steps to Reproduce

1. settings .aws config default region=ap-northeast-2 and zappa-settings.json's aws_region is us-east-1.
2. I do `zappa deploy stage` at local, and once time `zappa update stage` and `zappa certify stage` at local, and push gitlab and gitlab ci/cd's `zappa update stage` failed.

## Your Environment

* Zappa version used: 0.56.1
* Operating System and Python version: 3.9
* The output of `pip freeze`:
boto3==1.26.61
botocore==1.29.61
certifi==2022.12.7
zappa==0.56.1
(...)

And my local is mac book(intel) and, gitlab runner is ubuntu 20.04

If anyone knows how to solve the above error, I would be very thankful if you could help me! :)

PR #1217, opened 2023-02-07
Fix/Use aws_region in zappa_settings. Even if there is an os environment variable

kyaryunha:
## Description

Zappa README.md said 
```
"aws_region": "aws-region-name", // optional, uses region set in profile or environment variables if not set here
```

but if AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are written in environment variables, zappa sets aws_region to AWS_DEFAULT_REGION in environment variables even if aws_region is written in zappa_settings.json.

## GitHub Issues

The issue below is also resolved by export AWS_DEFAULT_REGION=us-east-1 ,
The problem was caused by zappa importing the AWS_DEFAULT_REGION of the environment variable, not the aws_region of zappa_settings.

https://github.com/zappa/Zappa/issues/1215


monkut:
@kyaryunha Thanks for the contribution.  

I think we can get this in the next release.
The linter is failing, you'll need to run `make black` / `make isort`.

kyaryunha:
10 failed files of CI/CD are displayed. (Like the picture below.)
So if I do "make black", 10 files are changed.

But my "Files Changed" is only 1 file.

Should I commit all 10 codes of "make black"?
Or should I just commit the 1 file I did?

스크린샷 2023-03-02 오후 2 56 09

souravjamwal77 (changes requested):
I think the error message could be like
`No aws_region found, set the aws_region in zappa_settings or Set the AWS_DEFAULT_REGION as env variable`

kyaryunha:
@souravjamwal77

Thanks for the nice phrase recommendation! I'm not good at English.

`,` instead of `-` seems to be the zappa LookupError pattern,
In the expression you recommended, I'll changed `,` to `-`
I'll edit it to `No aws_region found - set the aws_region in zappa_settings or Set the AWS_DEFAULT_REGION as env variable` :D

monkut (approved):

monkut (changes requested):
Sorry, can you add a testcase for this to assure that the newly added `LookUpError` is raised when the region is not defined?

kyaryunha:
Ok, I will add test code next week.

kyaryunha:
I tried to write a test code, but for some unknown reason, it didn't work.

I'll try again next week.

monkut (approved):

javulticat (changes requested):

Issue #1160, opened 2022-08-05
Add support for python 3.10


ppartarr:
Title says it all. Ubuntu 22 ships with python 3.10 by default so it would be nice to have zappa support 3.10

monkut:
While zappa now supports docker based lambda deployments, but aws doesn't not yet provide a python3.10 lambda runtime.
Once aws provides a 3.10 runtime, we'll work on getting support in.

monkut:
This issue covers the docker usecase:
https://github.com/zappa/Zappa/issues/1140

ipsbrittainm:
Any update on this? AWS now supports python3.10

dennybiasiolli:
Good question: the official aws base image supports Python 3.10 here https://github.com/aws/aws-lambda-base-images/blob/python3.10/Dockerfile.python3.10

sridhar562345:
@ipsbrittainm @dennybiasiolli

AWS didn't release the python3.10 runtime support for zip-based packages.
![Lambda runtime](https://user-images.githubusercontent.com/59284206/229089377-9538762c-7355-4a71-929e-1acb4b5290fa.png)

However, you can use python3.10 runtime with container-based deployments and it's already supported in Zappa.

gutomaia:
Any updates on this issue?

buko106:
https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

python3.10 on AWS lambda is READY ❗

monkut:
Added in 0.57.0 release, published to pypi 5/19, closing.

PR #1219, opened 2023-02-08
Update __init__.py

moriisaac:
Update to incorporate python 3.10.

### DIFF
--- a/zappa/__init__.py
+++ b/zappa/__init__.py
@@ -12,7 +12,7 @@ def running_in_docker() -> bool:
     return running_in_docker_flag
 
 
-SUPPORTED_VERSIONS = [(3, 7), (3, 8), (3, 9)]
+SUPPORTED_VERSIONS = [(3, 7), (3, 8), (3, 9),(3,10)]
 MINIMUM_SUPPORTED_MINOR_VERSION = 7
 
 if not running_in_docker() and sys.version_info[:2] not in SUPPORTED_VERSIONS:


katomaso:
+1

moriisaac:
😂😂😂

souravjamwal77:
Hi @katomaso and @moriisaac , if you're using docker with Zappa then you can use any version of python, even 2.7  or 3.11 too.

katomaso:
It's true that AWS does not support even 3.10 yet so this is a bit preliminary. But we could merge it still, no?

sridhar562345:
Can you tell us what exactly you are trying to achieve with this?

zappa/__init__.py
```diff
@@ -12,7 +12,7 @@ def running_in_docker() -> bool:
     return running_in_docker_flag
 
 
-SUPPORTED_VERSIONS = [(3, 7), (3, 8), (3, 9)]
+SUPPORTED_VERSIONS = [(3, 7), (3, 8), (3, 9),(3,10)]
```
kyaryunha, line 15: I don't know zappa well so I'm not sure if 3.10 is a supported version,
but I think it would be nice to have a blank space.

before
`SUPPORTED_VERSIONS = [(3, 7), (3, 8), (3, 9),(3,10)]`

after
`SUPPORTED_VERSIONS = [(3, 7), (3, 8), (3, 9), (3, 10)]`

monkut:
This value is used to define the lambda python runtime. As of 2023/3/1, python 3.9 is the latest supported runtime provided by aws.
https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

As @souravjamwal77 mentioned, if you want to use another python release, you can create and provide a docker image with your desired version.  I this case, the defined `SUPPORTED_VERSIONS` here is by-passed.

When AWS provides a 3.10, or 3.11 runtime we'll get a patch release out as soon as we can to support it.

monkut:
As mentioned AWS does not yet provide a python 3.10 (or 3.11) runtime.
If you want to go ahead and add preliminary support there are more updates and testcases need.

There may be more, but this needs to be updated:

https://github.com/zappa/Zappa/blob/master/zappa/utilities.py#L206-L216

And related testcase added:

https://github.com/zappa/Zappa/blob/master/tests/tests.py#L144-L169



Support for the 3.10 runtime is tracked in the following issue:
https://github.com/zappa/Zappa/issues/1160

hellno:
If anyone wants to follow the PR for python 3.10 support: https://github.com/zappa/Zappa/pull/1231
the issue is tracked in #1160

Issue #470, opened 2021-02-20
[Migrated] Feature/passing source ip


jneves:
Originally from: https://github.com/Miserlou/Zappa/issues/1249 by [danielwhatmuff](https://github.com/danielwhatmuff)

## Description
Passing env var for the $context.identity.sourceIp

http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html#context-variable-reference

## GitHub Issues
Fairly small change discussed in slack but let me know if you want an issue for further discussion.

monkut:
I don't see a problem adding the lambda context sourceIP, but currently the client IP should be available in "REMOTE_ADDR".

https://github.com/zappa/Zappa/blob/master/zappa/wsgi.py#L89-L96

PR #1228, opened 2023-04-12
Added support for passing the client source IP address from API Gateway.

nickovs:
## Description
This is a minor PR to check if the `requestContext` object provided by the API Gateway includes an `identity` object and that contains a `sourceIP` value representing the client IP address; if present then that value is made available to the application in the `CLIENT_SOURCE_IP` environment variable.

The PR also adds a test case for passing the client IP address.

## GitHub Issues

This addresses issue #470.


nickovs:
@monkut I've cleaned up the white space so that `make black` makes no changes. `isort` seemed happy so no imports were changed.

monkut:
Just re-reviewing this, and I believe that the client IP is available in "REMOTE_ADDR".

https://github.com/zappa/Zappa/blob/master/zappa/wsgi.py#L89-L96

This isn't the lambda context "sourceIp", but I suspect it should be the desired value.

nickovs:
@monkut Unfortunately the AWS application gateway doesn't properly set the `X-Forwarded-For` header, so the default deployment process for Zappa ends up with that header being empty. To see this try creating a new Zappa deployment for the following Flask app:
```
import os
from flask import Flask
app = Flask(__name__)

@app.route("/", methods=["GET", "POST"])
def lambda_handler(event=None, context=None):
    return "Remote IP is: {}".format(os.environ.get("REMOTE_ADDR"))

if __name__ == "__main__":
    app.run(debug=True)
```

When you access the page you will get: `Remote IP is: None`

nickovs:
Having said that, it might make sense to change the PR to put the result into `REMOTE_ADDR` instead of `CLIENT_SOURCE_IP`, so that app developers only have one place to look.

If you agree to that then I can make the change. Based on my observation so far, it looks we will never have both code paths executed in the same deployment, but if you have an opinion as to if the new code should override the old code or not in the event of a conflict then let me know.

monkut:
@nickovs

Ok, I checked and confirmed that I am using the REMOTE_ADDR value it a project.  
If it's not available by default from APIGW, I suspect that I configured the APIGW to allow the header through.

In that case can you update this PR to:

- Use `REMOTE_ADDR` for the client ip.
- Fallback to the context "sourceIp" if the `X-Forwarded-For` is unavailable.

PR #1235, opened 2023-04-25
Fix/Add something to exclude in zappa_settings, the default exclude does not work.

kyaryunha:
## Description

In the existing code, if there is no exclude in zappa_settings, boto3, botocore, etc. are excluded by default.

However, if the user adds something to exclude in zappa_settings,
Boto3 and botocore, which were excluded by default, are not excluded.

I found this out after the "AWS Lambda Error: Unzipped size must be smaller than 262144000 bytes" error in aws lambda.
When I looked it up, the botocore was very large at 77MB, and when installing zappa, the botocore was installed.

That was the butterfly effect of my adding something to zappa_setting's exclude.

Looking at the zappa code, if I **add something to exclude in zappa_settings, the default exclude does not work.**

The get syntax in python adds the argument at the end if there is no "exclude" in stage_config.
If present, the following argument is not added.

So, by solving this problem, I leave a pull request.

Before
```python
exclude = self.stage_config.get("exclude", ["boto3", "dateutil", "botocore", "s3transfer", "concurrent"])
```

After
```python
exclude = self.stage_config.get("exclude", [])
exclude += ["boto3", "dateutil", "botocore", "s3transfer", "concurrent"]
```

Additionally, can you confirm the reply I sent to the pull request I sent earlier? The lint problem doesn't seem to be happening in my code. (https://github.com/zappa/Zappa/pull/1217) 

And, thanks you about zappa, I can use aws deploy easier.

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -2440,7 +2440,8 @@ class ZappaCLI:
                 disable_progress=self.disable_progress,
             )
         else:
-            exclude = self.stage_config.get("exclude", ["boto3", "dateutil", "botocore", "s3transfer", "concurrent"])
+            exclude = self.stage_config.get("exclude", [])
+            exclude += ["boto3", "dateutil", "botocore", "s3transfer", "concurrent"]
 
             # Create a single zip that has the handler and application
             self.zip_path = self.zappa.create_lambda_zip(


sridhar562345:
Excluding boto3 and botocore by default is good, but there might be cases when the user has to use a newer version of boto3 that's not available on lambda. So I think there should be an option to disable default excludes.

souravjamwal77:
Hi @kyaryunha To exclude boto and botocore read below in the [advanced settings](https://github.com/zappa/Zappa#advanced-settings). `exclude_glob' might help you.
```
"exclude": ["file.gz", "tests"], // A list of filename patterns to exclude from the archive (see `fnmatch` module for patterns).
        
"exclude_glob": ["*.gz", "*.rar", "tests/**/*"], // A list of glob patterns to exclude from the archive. To exclude boto3 and botocore (available in an older version on Lambda), add "boto3*" and "botocore*".
```

kyaryunha:
@souravjamwal77 
Thanks for introducing `exclude_glob`, but that's not what I'm talking about.
To solve the problem, when I put other files in exclude, I just add boto3 to exclude, so I can do it with `exclude` only.

What I want to say is,
**I just excluded the `test` file, and _I didn't do anything about boto3_, so the exclude of boto3 should not be changed.**

This is the logic that needs to be separated.
If not, users need to look at the zappa code to know that the reason boto3 is included is because they added something else to exclude.

hellno:
Thank you @kyaryunha, I think this PR is a great isolated fix. 

@sridhar562345 can you create a separate issue with a description of the problem that you described above. 
If you see different ways of fixing it, we can discuss if any of them is the desired one or the implementer can decide which way to go. This will be a `good first issue` for someone to take.

hellno (approved):
🙌🏽🙌🏽🙌🏽

kyaryunha:
@hellno  
I added a new pull request that implements the other ideas above.
Read it and choose the better of the two, or if you have a better idea, go for it.

https://github.com/zappa/Zappa/pull/1242

hellno:
Lets take it in #1242 and continue the discussion there

PR #1238, opened 2023-05-01
lint: updating code style with `make black`

dennybiasiolli:

PR #1231, opened 2023-04-18
adding support for Python 3.10

dennybiasiolli:
Ref #1160, #1124

PR #1239, opened 2023-05-01
Nose to pytest

dennybiasiolli:
Migrating test suite from the old `nose` to the new `pytest`, keeping coverage results (aggregated) and duration display.

Inspired by @monkut in https://github.com/zappa/Zappa/pull/1231#issuecomment-1529344242

This PR goes after this one: https://github.com/zappa/Zappa/pull/1238
in order to preserve style formatting there and use it from now on.


monkut (approved):

monkut:
thanks for the hard work! 
Looks like `isort/black` needs to be run.

dennybiasiolli:
@monkut need to merge this one first, so we can fix the linting before everything else 😉

hellno (approved):
Amazing effort, thank you! 🙌🏽🙌🏽 LGTM

Issue #1241, opened 2023-05-04
Option to disable default excludes


sridhar562345:
Excluding boto3 and botocore by default is good, but there might be cases when the user has to use a newer version of boto3 that's not available on lambda. Currently, there is a workaround as below.

```py
exclude = self.stage_config.get("exclude", ["boto3", "dateutil", "botocore", "s3transfer", "concurrent"])
```

To use newer version of boto3, specify exclude in zappa_settings.json which will override the default excludes
```json
"exclude": ["file.txt"]
```

https://github.com/zappa/Zappa/pull/1235 will modify this snippet, but we won't be able to use the newer boto3 version at all.

```py
exclude = self.stage_config.get("exclude",
exclude += ["boto3", "dateutil", "botocore", "s3transfer", "concurrent"]
```
So having the option to disable default excludes would be better.

monkut:
After the 3.11 release, let's try to get this in.

monkut:
Just to note, you can currently disable the default excludes by setting, `excludes=[]`.

PR #1242, opened 2023-05-04
Fix/exclude add when initial zappa setting

kyaryunha:
## Description

My prev pull request is bad because it always disable boto. 

So I code a new idea about this mention.
https://github.com/zappa/Zappa/pull/1235#issuecomment-1521308031

## GitHub Issues
My Prev Pull Request: https://github.com/zappa/Zappa/pull/1235
Issue: https://github.com/zappa/Zappa/issues/1241


monkut (commented):
Can you add a testcase to confirm the results of this?

kyaryunha:
I wrote a test code to check if excludes are properly included in zappa_settings when I run zappa init.

It's the first time I've written test code in Python, so I'd appreciate it if you could check if the test code was written clearly.

In addition, when there is something in the exclude, please tell me if you need to write a test code to see if it is excluded well in the zip.

thank you!

kyaryunha:
Yes, this code conflicts with the old zappa_settings.json.
I think it would be nice to do it when updating a large version.

kyaryunha:
Ah.. ci failed... 
I'll think about it again and fix it this week.

zappa/cli.py
```diff
@@ -1871,6 +1871,7 @@ def init(self, settings_file="zappa_settings.json"):
                 "s3_bucket": bucket,
                 "runtime": get_venv_from_python_version(),
                 "project_name": self.get_project_name(),
+                "exclude": ["boto3", "dateutil", "botocore", "s3transfer", "concurrent"],
```
monkut, line 1874: Ok, so on initilization, the sample zappa_settings_file will include the excludes, but if excludes are removed from the settings, then none will be included by default, right?

If my understanding is correct, I think this is a great "fix" in how excludes work.

I believe this is mentioned in the comments, but this merge should wait until we're ready to issue another major release.

kyaryunha: Yes, that's right.
Since boto3 etc. are no longer excluded for users who do not have an exclude in zappa_settings,
So it's good to deploy on the next major release. (ex. 0.58.0)

PR #1258, opened 2023-07-03
Update Executing in Response to AWS Events section in README.md 

MirzaBaig715:
## Description
 
Update the README.md for [Executing in Response to AWS Events](https://github.com/zappa/Zappa#executing-in-response-to-aws-events). 

## GitHub Issues
[AWS Documenttion](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html) supports both FIFO and Standard queues to trigger lambda events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -619,7 +619,7 @@ Optionally you can add [SNS message filters](http://docs.aws.amazon.com/sns/late
        ]
 ```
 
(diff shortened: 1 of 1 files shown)


davidgolden:
Would be great to get this merged, since the current Zappa documentation is now out of date.

Old AWS Docs which state that only Standard Queues are supported by Lambda: https://web.archive.org/web/20190112132734/https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

New AWS Docs which state that Lambda supports both Standard and FIFO Queues: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

monkut (approved):

Issue #1229, opened 2023-04-14
Combined Log Format response times have wrong number of seconds/microseconds


nickovs:
## Context

Requests are logged with an `INFO` level log message that extends the standard Combined Log Format message with the time it took for the handler to process the request, nominally in seconds and microseconds:
```
[1681511864174] [INFO] 2023-04-14T22:37:44.174Z 755cd95f-b898-4a99-a247-cb1e97e2777e 1.2.3.4 - - [14/Apr/2023:22:37:44 +0000] "GET /main/ready/ HTTP/1.1" 200 5104 "https://some.app.net/auth/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.4 Safari/605.1.15" 0/3.615
```

In this example the execution time is represented by the `0/3.615` at the very end of the line.

The [code that implements the format](https://github.com/zappa/Zappa/blob/73ee393dd48500d321d65e92aba31a5fd83e93ab/zappa/utilities.py#L672-L688) expects to be passed a number of microseconds and this is divided by 1000000 to get the number of seconds. Unfortunately the [code that calls this](https://github.com/zappa/Zappa/blob/73ee393dd48500d321d65e92aba31a5fd83e93ab/zappa/handler.py#L621-L624) passes the response time in milliseconds, not microseconds. This leads to the number of seconds more or less always being 0.

## Expected Behavior
In the last part of the format, either the last number is in milliseconds and the second-to-last number is the last number divided by 1000, or the last number is in microseconds and the seconds-to-last number is that last number divided by 1000000.

## Actual Behavior
The last number is in milliseconds but the second-to-last number is that last number divided by 1000000, essentially making the second-to-last number be in a unit of thousands of seconds.

## Possible Fix
Either change the [code that calculates the time](https://github.com/zappa/Zappa/blob/73ee393dd48500d321d65e92aba31a5fd83e93ab/zappa/handler.py#L622) to compute the time in microseconds, or change the [code that computes the number of seconds](https://github.com/zappa/Zappa/blob/73ee393dd48500d321d65e92aba31a5fd83e93ab/zappa/utilities.py#L686) to divide the input by 1000.

[...]

PR #1265, opened 2023-08-16
:bug: fix response time improperly configured for micro-seconds.

monkut:
## Description

Microseconds were not properly being output into the log via the `common_log()` function.
This PR corrects the coversion of seconds to microseconds and clarifies the units used.

## GitHub Issues
https://github.com/zappa/Zappa/issues/1229


souravjamwal77 (approved):
A screenshot with the results was sent to @monkut. Approved

PR #1279, opened 2023-10-31
add keep_policies config to prevent Zappa from removing resource policy

aleontiev:
## Description

This is a small change to ignore certain policies from the pruning that happens in `Zappa._clear_policy`

The context is that my function is attached to a VPC/VPN/private ALB through a secondary configuration. It requires a policy that allows load-balancing, but the policy keeps getting cleared out by Zappa.

This change allows for a simple `keep_policies` configuration that can be set to a list or string for SID matching within the function policy, and is null/disabled by default.

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -117,6 +117,7 @@ class ZappaCLI:
     environment_variables = None
     authorizer = None
     xray_tracing = False
+    keep_policies = None
     aws_kms_key_arn = ""
     context_header_mappings = None
     additional_text_mimetypes = None
@@ -2295,6 +2296,7 @@ class ZappaCLI:
         self.aws_kms_key_arn = self.stage_config.get("aws_kms_key_arn", "")
         self.context_header_mappings = self.stage_config.get("context_header_mappings", {})
         self.xray_tracing = self.stage_config.get("xray_tracing", False)
+        self.keep_policies = self.stage_config.get('keep_policies', None)
         self.desired_role_arn = self.stage_config.get("role_arn")
         self.layers = self.stage_config.get("layers", None)
         self.additional_text_mimetypes = self.stage_config.get("additional_text_mimetypes", None)
@@ -2322,6 +2324,7 @@ class ZappaCLI:
             tags=self.tags,
             endpoint_urls=self.stage_config.get("aws_endpoint_urls", {}),
             xray_tracing=self.xray_tracing,
+            keep_policies=self.keep_policies
         )
 
         for setting in CUSTOM_SETTINGS:
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -2752,7 +2752,7 @@ class Zappa:
                 for s in statement:
                     sid = s['Sid']
                     if self.keep_policies:
-                        if not (self.keep_policies == '*' or (isinstance(self.keep_policies, list) and sid in self.keep_policies) or (isinstance(self.keep_policies, str) and self.keep_policies in sid)):
+                        if self.keep_policies == '*' or (isinstance(self.keep_policies, list) and sid in self.keep_policies) or (isinstance(self.keep_policies, str) and self.keep_policies in sid):
                             # * = keep all
                             # string = keep sids that contain the value
                             # list = keep sids that are in the list (no wildcards)
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -286,6 +286,7 @@ class Zappa:
         tags=(),
         endpoint_urls={},
         xray_tracing=False,
+        keep_policies=None
     ):
         """
         Instantiate this new Zappa instance, loading any custom credentials if necessary.
@@ -335,6 +336,7 @@ class Zappa:
 
         self.endpoint_urls = endpoint_urls
         self.xray_tracing = xray_tracing
+        self.keep_policies = keep_policies
 
         # Some common invocations, such as DB migrations,
         # can take longer than the default.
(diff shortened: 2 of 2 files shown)


monkut:
Can you raise an issue describing the problem addressed here?

javulticat:
> ## Description
> 
> This is a small change to ignore certain policies from the pruning that happens in `Zappa._clear_policy`
> 
> The context is that my function is attached to a VPC/VPN/private ALB through a secondary configuration. It requires a policy that allows load-balancing, but the policy keeps getting cleared out by Zappa.
> 
> This change allows for a simple `keep_policies` configuration that can be set to a list or string for SID matching within the function policy, and is null/disabled by default.

I am _fairly_ confident that there are probably existing Zappa settings that can do some, if not basically all, of what you are trying to accomplish with this PR.  It's not 100% clear to me what exact behavior you need to have (or to avoid?) from Zappa, but if you read through the [**README**](https://github.com/zappa/Zappa#readme), especially once it starts talking about Zappa's [**Advanced Settings**](https://github.com/zappa/Zappa#advanced-settings) and various specific use cases for them (with many of those use cases also linking to some useful tutorials some of our users have made), I think there's a good chance you'll find what you need somewhere in the documentation.

In particular, there are several examples under the [**Using Custom AWS IAM Roles and Policies**](https://github.com/zappa/Zappa#using-custom-aws-iam-roles-and-policies) that stand out in my mind as possibly being able to help with what it seems like you might be trying to do with this PR:

1. [**Custom AWS IAM Roles and Policies for Deployment**](https://github.com/zappa/Zappa#custom-aws-iam-roles-and-policies-for-deployment): If you are encountering issues with the default Zappa IAM role not being able to deploy your application successfully, take a look at this example and consider using the `profile_name` setting.  A profile can be tied to an IAM user/role that can be granted whatever specific policies/permissions you need so Zappa is able to deploy your application successfully in your specific environment.
2. [**Custom AWS IAM Roles and Policies for Execution**](https://github.com/zappa/Zappa#custom-aws-iam-roles-and-policies-for-execution): If the problem is that, once deployed, Zappa seems to be unable to run your application, then perhaps take a look at this example.  By setting `"manage_roles": False` and then setting either `role_name` or `role_arn` to a custom role name/arn you've set up without the help of Zappa that you can grant whatever specific policies/permissions that the role needs to be able to successfully run your application in your environment.
    - I'll note that it sounds like using the existing settings `{"manage_roles": False, "role_name": "MyIAMPolicyRole"}` would do something _very_ similar to what you are trying to accomplish with the new setting you suggest adding in this PR (e.g., `{"keep_policies": "MyIAMPolicyRole"}`).
3. Furthermore, you can see just below the [previous example](https://github.com/zappa/Zappa#custom-aws-iam-roles-and-policies-for-execution) that you can also use the `extra_permissions` setting to essentially explicitly add a list of additional JSON IAM policies containing whatever permissions you'd like for the Zappa IAM role to have.  This can be used as an alternative to provisioning and managing a custom IAM role, as this will have Zappa still manage the IAM role, but just with whatever extra policies/permissions you give it in your Zappa settings file.
4. Also, as a kind of compromise between the relative extremes of telling Zappa to stop managing IAM at all and having to maintain the Zappa execution role yourself, or having to put all of the custom IAM policies/permissions you need the Zappa IAM role to have in one giant `extra_permissions` block in your settings file, I believe a third option is that you could also choose to use the `assume_policy` and/or `attach_policy` settings, which you'd do by defining them as strings containing the filename (and qualified path, if necessary) of a valid JSON IAM Policy file (ex. `"assume_policy"`: "my_very_long_policy_file.json"), which can then allow the policies/permissions needed by the Zappa IAM role to live in its own JSON IAM Policy file, which you can then have the Zappa IAM role assume or attach, respectively, to gain the permissions it needs when executing your application in your environment.

And, if it helps, Zappa can also be configured to run within a VPC using some other settings you may want to take a look at.  For starters, check out the `vpc_config` dictionary setting near the very bottom of the list of [**Advanced Settings**](https://github.com/zappa/Zappa#advanced-settings), and then you can read a bit more about running Zappa within a VPC here: [**Running Tasks in a VPC**](https://github.com/zappa/Zappa#running-tasks-in-a-vpc).  That section should also contain links to external tutorials that go into more depth on the subject.

Speaking of tutorials, that reminds me: don't forget to check out our [**Zappa Guides**](https://github.com/zappa/Zappa#zappa-guides) section of the [**README**](https://github.com/zappa/Zappa#readme) - several cover topics around deploying and running Zappa within a VPC and other topics you might find helpful for your specific situation.

You may also want to check out our section on how to configure API Gateway to make your Zappa application _only_ be accessible from within your private VPC, if that's a use case relevant to you.  See: [**Endpoint Configuration**](https://github.com/zappa/Zappa#endpoint-configuration)

Lastly, I'll also just mention that Zappa has some ALB support out-of-the-box as well, including supporting ALBs running within VPCs.  Similar to the `vpc_config` I mentioned earlier, take a look at the `alb_vpc_config` dictionary setting near the very top of the list of [**Advanced Settings**](https://github.com/zappa/Zappa#advanced-settings) (and don't forget about the `alb_enabled` setting either!), and then you can read a bit more about running Zappa behind an ALB: [**Application Load Balancer Event Source**](https://github.com/zappa/Zappa#application-load-balancer-event-source).  I think that's yet another section that contains some links that go into more depth on the subject.

With sufficient targeted Zappa research under your belt, I would bet that you will be able to accomplish what you are trying to without needing to open a PR to add any additional settings to Zappa. But, then again, perhaps I have just not fully understood what you are trying to accomplish... 😕

That's why, if you do the research and still feel like you are no closer to solving your problem on your own, then I strongly agree with and bump @monkut's [_wise suggestion_](https://github.com/zappa/Zappa/pull/1279#issuecomment-1794039835) to please create a new [**Issue**](https://github.com/zappa/Zappa/issues/new), which will provide you with a dedicated, more visible space designed for discussion to happen among the Zappa community.  You'll be able to share more on the specifics of the problems you are facing, and that will help us all to get on the same page 🙌

And, who knows, you might even get lucky and have another member of the Zappa community who has dealt with the same issue before might show up and tell you how he figured out how to solve it.  That's certainly something I've seen happen a few times.

Anyway, I really hope something in the documentation or a tutorial will help the solution to your problem click for you, but, if not, I look forward to reading the Issue you open about it and learning in more detail about why the current Zappa implementation seems to be incompatible with achieving your goals under your specific set of circumstances, and collaborating on what would be the ideal way to solve whatever the root of that problem ends up being 🤔

Good luck!

Issue #1282, opened 2023-11-06
Adding support for `pre-commit` hooks


javulticat:
Zappa is a project that began development over 8 years ago and still relies heavily on a `Makefile` (which are pretty outdated in Python projects) and human memory to do local development.  With our current setup, I've probably had to remind users to run `make black` or `make isort` hundreds of times in Zappa PRs at this point.

There's a better way.  [`pre-commit`](https://pre-commit.com/) is now more-or-less the industry standard to ensure that things developers need to run before pushing a commit are run automatically, without the developer needing to remember to do so.  I haven't worked a job in probably half a decade that didn't make use of `pre-commit` in their repos.  Setting up `pre-commit` is on the checklist of "first things to do" when setting up a personal repo or a repo for my engineering team.

The proverbial straw that broke the camel's back for me today is that I was making an extremely minor change to our README, and the tests complained I needed to run [`doctoc`](https://github.com/thlorenz/doctoc) before I could merge my changes.  Now, the use of and need for `doctoc` isn't documented anywhere in the Zappa repo AFAIK, but that's a separate problem for another day.

The main issue I had with that is that `doctoc` is a Node.js package that wanted me to install it with `npm`.  I didn't have Node.js or `npm` installed in the environment I was working in.  And, imo, it's just **absolute insanity** that we require contributors to install Node.js, a Node package manager, and a specific node package just to be able to contribute to _the damn README_ of our **Python** project.

Now, fortunately, like most tools found in the civilized world, [`doctoc` offers a `pre-commit` hook](https://github.com/thlorenz/doctoc#usage-as-a-git-hook), which ameliorates the problem, as `pre-commit` automatically takes care of configuring the necessary environment to run any `pre-commit` hook you can find.  So, no more need for a user wanting to change a line in the README to have to install Node.js, `npm`, and `doctoc` on their machine; only `pre-commit` is needed, which can be installed via `pip` just like any other Python package.

In fact, every linting tool we use (and several others we probably should be using): `black`, `isort`, `flake8`, `mypy`, etc., all have useful `pre-commit` hooks.

So, I configured and successfully tested using `pre-commit` to run `doctoc` on Zappa for me, so I'll be adding the rest of our linting tools as well to a `pre-commit` configuration for the repo.  I'll also see about the feasibility of removing/reducing the need for the `Makefile`, though will likely do so through a separate PR.

TL;DR: `pre-commit` is coming soon to a Zappa near you!

javulticat:
Done in #1284

PR #1209, opened 2022-12-11
PR-1123 Add Support for Graviton 2 / ARM Architecture (conflict fix)

gaborschulz:
## Description
[PR-1123 Add Support for Graviton 2 / ARM Architecture](https://github.com/zappa/Zappa/pull/1123) has merge conflicts. This PR intends to fix those conflicts.

## GitHub Issues

Resolves: [#1051](https://github.com/zappa/Zappa/issues/1051)

PR #1284, opened 2023-11-07
Configures 'pre-commit' hooks for Zappa's development tools

javulticat:
Adds [`pre-commit`](https://pre-commit.com/) hooks for all of the major linting/formatting tools currently used in Zappa development.

Two choices I intentionally made to keep the scope of this PR small, low-risk, and easy to review:

1. I specifically tried to keep the behavior of the `pre-commit` hooks to remain as close as possible to the currently defined behavior of the tools in our `MAKEFILE`, even though, in many cases, that behavior is sub-optimal, or even, in some cases, entirely useless (for example, our `MAKEFILE` has `flake8` currently running with the `--exit-zero` arg always enabled, meaning that our CI always thinks it succeeds, despite the fact that `flake8` does, in fact, not actually _ever_ succeed when currently run on our repo and in fact, ends up finding a number of errors that this `MAKEFILE` configuration has hidden for... likely years at this point).  I plan on properly configuring our dev tools to actually provide the value they are meant to in follow up PRs.
2. I also specifically decided to not make any changes to the actual Python code powering Zappa in this PR.  The _only_ changes this PR makes to `.py` files are adding comments that instruct `mypy` to ignore certain lines that don't currently pass `mypy`'s checks.  Again, modernizing the codebase to better conform to all of the best practices imposed by tools like `flake8` and `mypy` can wind up being massive undertakings on big, old, largely untyped codebases like ours that have seemingly never prioritized best practices too highly... So, I wanted to leave any code changes for the sake of pleasing our linting tools (and, thereby hopefully making our code much more readable and easy to follow) to be done via separate PR(s) as well so this one could be safely merged and begin making our contributors' lives easier sooner rather than later.

For more, see Issue #1282, which will be resolved by this PR.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,45 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.5.0
+    hooks:
+    -   id: check-yaml
+    -   id: name-tests-test
+        args: [--unittest]
+        exclude: ^tests/utils.py
+-   repo: https://github.com/pycqa/isort
+    rev: 5.12.0
+    hooks:
+    -   id: isort
+        args: [--profile=black, --gitignore]
+-   repo: https://github.com/psf/black
+    rev: 23.10.1
+    hooks:
+    -   id: black
+        args: [--line-length=127]
+-   repo: https://github.com/PyCQA/flake8
+    rev: 6.1.0
+    hooks:
+    -   id: flake8
+        args:
+        -   --count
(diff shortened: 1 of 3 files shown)


monkut (approved):

monkut:
@gaborschulz introduced pre-commit in the pull request below.
It has a conflict and hasn't been merged in yet.  

This looks good, I'm ok with getting this in, and the requesting conflict resolution when this is in.

https://github.com/zappa/Zappa/pull/1209

PR #1287, opened 2023-12-18
Fix wheels regex

martimarkov:
Fix so that more wheels can be found.
E.g. `lxml-4.9.3-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.manylinux_2_24_x86_64.whl`

## Description
Change the regex in order to capture more wheels from PyPi or local cache. 

## GitHub Issues
https://github.com/zappa/Zappa/issues/1288


hellno (approved):

monkut (changes requested):
Can you make a test case for this?

PR #1300, opened 2024-02-03
Enable live runs of repo maintenence job via manual trigger

javulticat:
## Description
Following up on the new automated repo maintenance job added in #1296 to help with handling ancient/inactive/abandoned issues/PRs, it has been [successfully completing dry-runs as expected for a while now](https://github.com/zappa/Zappa/actions/workflows/maintenance.yml) - it has run without issue since #1298 (which fixed a typo that was causing the job to fail) was merged ~2 weeks ago, and it's currently at ~150 consecutive successful runs (and counting).

This PR represents the next step toward enabling this bit of repo maintenance automation: dry-runs are being disabled, and live runs will begin.  However, to err on the side of caution, this PR is also disabling the job's schedule which causes GitHub to automatically run it every 2 hours.  So, just in case a live run causes an unexpected problem, the job won't keep running every 2 hours until a PR is able to be merged.

Since the job lacks automated scheduling for now, I will manually execute the job and verify each of its runs to make sure the job worked as expected before allowing it to run again.  Once the job has successfully completed enough live runs that we can be highly confident that it is consistently performing as expected, the final step will be to re-enable the schedule for this job in a subsequent PR, allowing it to run automatically once again.

### Blocked by

 - [x] #1301


.github/workflows/maintenance.yml
```diff
@@ -2,8 +2,10 @@ name: Perform automated repository maintenance
 
 on:
```
monkut, line 3: Not sure an empty section is valid.
It may be better to comment out the whole section block.

javulticat: The section is not actually empty 🙃, [the next line](https://github.com/zappa/Zappa/pull/1300/files#diff-6eb5da11d7badb128a54d987e293ca9eec3cbb461d01a0abeae2d7a09675aefbL4) (`workflow_dispatch`) is part of this `on` section, and these two lines were intentionally _not_ commented-out because `workflow_dispatch` is the "event trigger" that allows a workflow to be run manually[^1].

The evidence that these lines work as expected can already be seen when viewing [the workflow from the Actions tab](https://github.com/zappa/Zappa/actions/workflows/maintenance.yml), where we can see that the **`Run workflow`** button which allows the workflow to be run manually[^2] is successfully appearing, along with a message that confirms that `workflow_dispatch` is "wired" correctly, allowing the button to appear:
>This workflow has a `workflow_dispatch` event trigger.

image

[^1]: https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow
[^2]: https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow#running-a-workflow

javulticat:
Decided to merge this as it had been stalled.  I've since tested and confirmed that this has had the expected effects:

- [x] Manual runs can be triggered as expected
image

- [x] Manually triggered dry-run ran successfully
image

- [x] Manually triggered live run ran successfully...
image

    - [x] ...which left comments/labels as expected, starting from the oldest open issues/PRs
image

PR #1316, opened 2024-03-29
Proper upload_to_s3 error messaging

### LINKED ISSUE #1315: Zappa tries to create new S3 bucket when it can't access the already existing one
## Context

If a user calls `zappa update {{stage}}` without the AWS `s3:ListBucket` permission the `upload_to_s3` function in `core.py` it tries to create a new bucket instead of alerting the person they cannot access the bucket.

## Expected Behavior

The `zappa update {{stage}}` call should error out and alert the user they cannot access the bucket they are trying to.

## Actual Behavior

The `zappa update {{stage}}` continues and creates a new s3 bucket or, in my case errors out because the does not have permission to create a new S3 bucket.

## Steps to Reproduce

1. Successfully deploy a zappa project
2. Create an AWS user or role with the following permissions.
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": "logs:UpdateLogDelivery",
            "Resource": "*"
        },
        {
            "Sid": "VisualEditor1",
            "Effect": "Allow",
            "Action": [
                "lambda:UpdateFunctionCode",
                "s3:PutObject",
                "s3:GetObject",
                "lambda:GetFunction"
            ],
            "Resource": [
                "arn:aws:lambda:*:<account>:function:*",
                "arn:aws:s3:::*/*"
            ]
        }
    ]
}
```
3. Call `zappa update {{stage}}`

## Your Environment

* Zappa version used: 0.58.0
* Operating System and Python version: MacOS Python 3.11

lmuther8:
## Description
 
**New Behavior**
The zappa update {{stage}} call should error out and alert the user they cannot access the bucket they are trying to.

**Old Behavior**
The zappa update {{stage}} continues and creates a new s3 bucket or errors out because the does not have permission to create a new S3 bucket.

## GitHub Issues

Closes #1315


monkut (changes requested):
Can you add a testcase for this?

PR #1311, opened 2024-03-20
Provide a fix for the module importer, deprecated `find_module` in Python 3.12

dennybiasiolli:
## Description

Deprecated `find_module` in Python 3.12, triggered when executing pre or post build scripts.
This PR fixes the issue

Ref: #1291, #1292

Issue #1291, opened 2024-01-06
About Python 3.12 support 


jagadeesh32:
Hello, Can you please tell,  when will expect zappa support with python 3.12?

javulticat:
Hi @jagadeesh32, PR #1292 added support for Py3.12, so the next version of Zappa we release will include this support.

If necessary, you can also install Zappa directly from our `master` branch (or from the specific commit that added Py3.12 support: https://github.com/zappa/Zappa/commit/77689e2a992f19f70c1372c3a48f930e45337431) in the interim before we publish our next release to obtain Py3.12 support immediately.  For example, with `pip`:

```
# Install from branch name
pip install git+https://github.com/zappa/Zappa.git@master

# Install from commit hash
pip install git+https://github.com/zappa/Zappa.git@77689e2a992f19f70c1372c3a48f930e45337431
```

PR #1320, opened 2024-04-03
'find_module' removed in Py3.12; use 'find_spec' instead

javulticat:
## Description

Shout out to @dennybiasiolli for pointing out this issue in #1311! 🙏  However, instead of the branching logic that creates separate codepaths for Py3.12 found in that PR, I believe a simpler solution is possible:

Since the [`find_spec` impl is backwards compatible with `find_module` and has been available since Py3.4](https://docs.python.org/3.11/library/importlib.html#importlib.abc.MetaPathFinder.find_module), it should work as a drop-in replacement on all of the Python runtimes Zappa currently supports (Py3.8 - Py3.12), so branching logic shouldn't be necessary.

Note: I intentionally kept this PR as small as possible in order to help expedite it, as it's one of the last remaining blockers for [our next release](https://github.com/zappa/Zappa/milestone/1) that will bring long-awaited support for Py3.12 to Zappa.  That said, further improvements can definitely be made to Zappa's aging CLI impl, such as following DRY principles to consolidate the repetitive lines of code seen here.  However, IMO, code cleanup is low priority such that it shouldn't block our next release; instead, it can be considered for inclusion during planning for subsequent Zappa release cycles.

### Related

Supersedes #1311
Relates to #1291


dennybiasiolli (commented):
Nice! And without an "if" in the code, makes perfect sense 😉

PR #1325, opened 2024-04-08
extend SQS event handler

eviltwin:
## Description
- report_batch_item_failures for partial batch processing
- maximum_concurrency to provide limits to SQS event source scaling
- remove outdated documentation about FIFO queues being unsupported

### DIFF
--- a/README.md
+++ b/README.md
@@ -620,7 +620,7 @@ Optionally you can add [SNS message filters](http://docs.aws.amazon.com/sns/late
        ]
 ```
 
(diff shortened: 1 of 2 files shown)


eviltwin:
It was quite a small extension of the existing SQS handler, so I wasn't sure how much documentation would need adding for it... I'm happy to elaborate more on what I put, but the existing options didn't come with much explainer so I wasn't sure if that was a deliberate choice or not.

I corrected the FIFO docs part mostly in passing (I have confirmed that a FIFO queue works just fine).

monkut (commented):
It looks like we don't have good test coverage around this, but can you try to add testcases to cover this addition?

It appears we have minor coverage in `test_placebo.py` has event_source related testcases.

eviltwin:
@monkut thanks, I'd looked for the test coverage when SQS events were originally implemented but I couldn't find them. I'll try to find the time in the next week to add coverage in the places you indicated :)

Issue #954, opened 2021-03-11
(Discussion) Keep using kappa library


Yaronn44:
The kappa library used for the events sources seems to be have been totally  abandoned. The last release is from february 2017...
AWS and Boto3 have a lot new functionalities and it seems that currently we just override the kappa main classes to add those new functionalities when they comes up (e.g the `ExtendedSnsEventSource`).

Wouldn't it be more maintanable to create a whole new class in the Zappa project for this only purpose ?

There is nothing really hard in this, it just need some time to think about making it really generic, robust and easily maintainable using only Boto3.
All of the main code for event sources is in the `utilities.py` so there is no huge changes accross the project.

Any other ideas are welcome, and I can take some of my time to start this refacto if you think it can be a good thing to do.

Related issues :
#413
#504 - #545
#512 - #668 - #809
#557
#718
#761 - #762
#817
#854 
#861
#907
#958

techdragon:
As an interim approach, perhaps vendoring kappa into the project as a submodule, to get full control and be able to make changes, and then progressively replacing the code would be a good way to minimise the amount of change this might cause?

whatnick:
This issue is affecting my use cases directly around event driven data processing for IoT sensors. I am happy to pitch in with PR's if the core contributors can come up with work breakdown structure and specific issues to address. I am all for vendoring this library and overriding sections to the more maintained boto3 equivalents.

cometurrata:
What about using terraform as it is well maintained ? 
In the [readme](https://github.com/zappa/Zappa#readme) there is an example : 

https://github.com/dpetzold/terraform-aws-zappa

monkut:
Reviewing issues, and this looks like something that should be addressed.

I don't think the `kappa` dependency should be necessary, and agree that the functionality should be pulled in to zappa, ideally without adding a new dependency.  

Most likely the code would be re-written with boto3 (an existing dependency of zappa).


Related code.
https://github.com/zappa/Zappa/blob/master/zappa/utilities.py#L219-L244

monkut:
In my mind, the first step to migrate away from kappa is to encapsulate the kappa components into a new zappa internal "infrastructure" api.

May be make a new "zappa.infrastructure" package with an `InfrastructureManager()` object to abstract out the kappa code.
Once that's done, we can then work to replace the kappa components from under the `InfrastructureManager()` api.

monkut:
here

monkut:
Handled in PR #1386 will be included in the next release.

[...]

PR #1334, opened 2024-05-05
remove kappa dependency and add kappa as a module [DON'T MERGE YET]

souravjamwal77:
This commit is the first step towards removing Kappa as a dependency of Zappa, so that we can add latest and more AWS features and not be limited by the features offered by Kappa.
Issue: #954

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -25,7 +25,6 @@ boto3 = ">=1.17.28"
 durationpy = "*"
 hjson = "*"
 jmespath = "*"
-kappa = "==0.6.0"
 pip = ">=9.0.1"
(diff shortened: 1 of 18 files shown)


souravjamwal77:
@lmuther8 Sure. Let's connect on the weekend. I was planning to do this in May but got occupied with something else.

monkut:
@souravjamwal77 

Sorry, the initial step wasn't to bring the `kappa` source code into zappa, but rather *wrap* it's usage so that `kappa` could be refactored out eventually.

souravjamwal77:
@monkut You can send me the steps on Slack. I'm little occupied but this week will atleast get this started.

Issue #1368, opened 2025-04-09
Support Lambda qualifier for invoke and manage CLI commands


oliviersels:
## Description

The Zappa CLI `invoke` and `manage` commands currently always execute on the $LATEST deployed version of the Lambda function. We should support specifying a specific version or alias to invoke in the CLI using an optional --qualifier argument.

```bash
zappa manage dev migrate --qualifier 1234
```

Specifying the correct version is especially important when performing for example a django migrate right after deploying a new version of the app. We've seen a race condition in our CI/CD pipeline where zappa did not invoke the recently deployed version of the app to run the migrations, but used the older version. Obviously this resulted in unwanted behavior.

Having this option gives the possibility to manually specify the newly deployed version.

I'll open a PR with the functionality.

PR #1369, opened 2025-04-09
Lambda invocation qualifier and client-context option (#1368)

oliviersels:
## Description
The CLI `invoke` and `manage` commands now have a --qualifier option that allows invoking a specific Lambda function version or alias.

The --client-context option was also added to allow sending a ClientContext to the Lambda function.

## GitHub Issues
#1368


monkut:
errors fixed in:

https://github.com/zappa/Zappa/pull/1394

closing

PR #1374, opened 2025-04-24
Remove depreciated python3.8 runtime support/Increment version -> 0.60.0

monkut:
## Description

AWS Lambda no longer supports the `python38` runtime, removing from zappa.

## Changes

🔥 remove python 3.8 support (depreciated Oct 14, 2024) 
🔧 change os.sys.platform (returning 'win32') usage to use sys.platform
:arrow_up: bump version 0.59.0 -> 0.60.0 in prep for release


tests/utils.py
```diff
@@ -4,6 +4,7 @@
 from collections import namedtuple
 from contextlib import contextmanager
 from io import IOBase as file
+from pathlib import Path
```
maruryota, line 7: I think this import statement is unnecessary.

maruryota (approved):
Almost no problem, but we felt it would be better to correct one commented area.

maruryota (approved):
LGTM

PR #1378, opened 2025-05-20
Incrementing release version to 0.60.1

monkut:
## Description

Since function url is not "fully" supported, incrementing minor ver only


goya813 (approved):
LGTM

PR #1390, opened 2025-09-17
Fix API Gateway v2 double path prefix issue

### LINKED ISSUE #1389: Issue: API Gateway v2 events cause double path prefix in WSGI routing
## Problem
When using API Gateway v2 events with a stage (e.g., `dev`), requests to `{apigatewayurl}/{stage}/collection` result in incorrect routing.

## What happens
- Request: `https://api.execute-api.us-east-1.amazonaws.com/dev/collection`
- `rawPath` in event: `/dev/collection` (includes stage)
- `script_name` set to: `/dev`
- Final route becomes: `/dev/dev/collection` ❌

## Expected behavior
- Final route should be: `/dev/collection` ✅

## Root cause
The `create_wsgi_request` function in `zappa/wsgi.py` doesn't properly strip the stage from `rawPath` when it's included in API Gateway v2 events.

## Current code (lines 125-130)
```python
if script_name in path_info:
    environ["PATH_INFO"].replace(script_name, "")
```

## Issues
1. `replace()` doesn't modify the original string
2. Checks if `script_name` is anywhere in path instead of at the beginning

## Impact
- Routes fail to match expected patterns
- Affects all API Gateway v2 deployments with stages
- Works fine for v1 events and custom domains

## Environment
- Zappa version: latest
- API Gateway: v2 events
- Python: 3.x

## Reproduction
Deploy with a stage and make requests to `{gateway_url}/{stage}/your-endpoint` - routes will have double stage prefix.

0xrohitgarg:
## Problem
API Gateway v2 events include the stage in `rawPath`, causing double path prefix in WSGI routing. Requests to `{gateway_url}/{stage}/endpoint` result in routes like `/stage/stage/endpoint` instead of `/stage/endpoint`.

## Solution
Replace flawed `replace()` logic with proper `startswith()` check to correctly strip `script_name` from `PATH_INFO` when present.

## Changes
- Fix path handling in `create_wsgi_request()` function
- Use `startswith()` instead of `in` check
- Properly assign the modified path instead of using `replace()`

## Testing
- Verified fix resolves double prefix issue
- Confirmed backward compatibility with existing functionality
- Tested various edge cases (no stage, different stages, etc.)

Fixes #1389

### DIFF
--- a/zappa/wsgi.py
+++ b/zappa/wsgi.py
@@ -126,8 +126,8 @@ def create_wsgi_request(
         environ["SCRIPT_NAME"] = script_name
         path_info = environ["PATH_INFO"]
 
-        if script_name in path_info:
-            environ["PATH_INFO"].replace(script_name, "")
+        if path_info.startswith(script_name):
+            environ["PATH_INFO"] = path_info[len(script_name):]
 
     if remote_user:
         environ["REMOTE_USER"] = remote_user


0xrohitgarg:
Hi @monkut, could you please take a look at this PR when you get a chance?

monkut (commented):
Can you add a testcase to define/reproduce the issue?

monkut:
issue handled, testcases added, and re-implemented in:
https://github.com/zappa/Zappa/pull/1391

Closing.

PR #1418, opened 2026-01-26
Feature/aws managed capacity

### LINKED ISSUE #1420: [Feature] Support Lambda managed instances capacity providers
https://docs.aws.amazon.com/lambda/latest/dg/lambda-managed-instances-capacity-providers.html

AWS Lambda recently introduced capacity providers to run Lambda functions on Amazon EC2 instances, providing flexibility in compute capacity and cost optimization for steady workloads.

wangsha:
## Description

closes #1420 
[AWS Lambda Managed Instances Capacity Providers](https://docs.aws.amazon.com/lambda/latest/dg/lambda-managed-instances-capacity-providers.html) enables running Lambda functions on your own EC2 instances, giving tighter control over compute capacity, networking, and cost. Capacity Providers define the instance types and scaling behavior used to execute your functions.

This PR updates the project to support (and document) configuring Lambda Managed Instances via Capacity Providers, making it possible to opt into managed instance execution when you need predictable capacity or specific instance characteristics.

## Key changes

Adds/updates configuration and documentation for defining a Lambda Capacity Provider.
Wires the new settings into the deployment flow (no behavior change unless explicitly enabled).


monkut (commented):
## PR Review: Feature/AWS Managed Capacity (#1418)

Thanks for the contribution! This adds support for [AWS Lambda Managed Instances Capacity Providers](https://docs.aws.amazon.com/lambda/latest/dg/lambda-managed-instances-capacity-providers.html) — a useful feature for users needing predictable capacity on their own EC2 instances.

Overall the feature is well-structured with good test coverage for the new functionality. However, there are several issues that need to be addressed before merging, ranging from bugs that will cause runtime errors to a breaking behavioral change.

---

### Critical Issues

**1. `_clear_policy()` regression — will delete ALL Lambda permissions (zappa/core.py)**

This is the most concerning change. The current code in `master` only removes CloudWatch Events permissions (created by `schedule_events`), intentionally preserving API Gateway, SNS, S3, and other service permissions. The PR changes this to delete *every* permission statement except `FunctionURLAllowPublicAccess`.

This will break any deployment that uses API Gateway, SNS triggers, S3 event notifications, or any other service-based permission. The original filter (`principal.get("Service") == "events.amazonaws.com"`) was deliberate and well-documented.

The fix should preserve the original CloudWatch Events filter while also skipping `FunctionURLAllowPublicAccess`:

```python
if s["Sid"] in ["FunctionURLAllowPublicAccess"]:
    continue
# Only remove CloudWatch Events permissions (created by schedule_events)
principal = s.get("Principal", {})
if isinstance(principal, dict) and principal.get("Service") == "events.amazonaws.com":
    delete_response = ...
```

**2. Missing `return` after "no function url" check — will IndexError (zappa/core.py)**

Both `update_lambda_function_url_domains()` (line ~1854) and `undeploy_function_url_custom_domain()` (line ~2783) log "no function url configured" but don't actually `return`. Execution continues and immediately hits `response["FunctionUrlConfigs"][0]` which raises `IndexError` on an empty list.

```python
# Both methods need:
if not response.get("FunctionUrlConfigs", []):
    logger.info("no function url configured on lambda, skip...")
    return  # <-- missing
```

**3. `deploy_lambda_function_url()` return value mismatch (zappa/cli.py:958)**

In `deploy()`, the PR assigns `endpoint_url = self.zappa.deploy_lambda_function_url(**kwargs)` but `deploy_lambda_function_url` returns the full response dict, not a URL string. This will cause `touch_endpoint()` to fail. Should be:
```python
endpoint_url = self.zappa.deploy_lambda_function_url(**kwargs)["FunctionUrl"]
```

**4. Inconsistent return types between `deploy_lambda_function_url` and `update_lambda_function_url`**

`deploy_lambda_function_url` returns the full response dict. `update_lambda_function_url` returns `response["FunctionUrl"]` (a string) in the update path but falls through to `deploy_lambda_function_url` (returning a dict) in the create path. Callers can't rely on the return type. Normalize both to return the URL string.

**5. `update_domain_name()` — `res` undefined when `self.apigateway` is falsy (zappa/core.py:~3353)**

The method wraps the `apigateway_client.update_domain_name()` call in `if self.apigateway:` but then unconditionally returns `res`. If `self.apigateway` is falsy, `res` is never defined → `NameError`.

**6. `Architectures` added to `update_function_code` kwargs (zappa/core.py:1318)**

`Architectures` is not a valid parameter for the `update_function_code` API. It can only be set via `create_function` or `update_function_configuration`. This will cause an API error on every update.

**7. Python 3.9 compatibility — `str | None` type hint (zappa/core.py:1623)**

The `str | None` union syntax requires Python 3.10+. The project supports Python 3.9. Use `Optional[str]` instead.

---

### Moderate Issues

**8. `touch_url` may be uninitialized (zappa/cli.py)**

In the `update()` method, `touch_url = endpoint_url` is set, but `endpoint_url` itself may not be set if the function URL update path returns inconsistent types (see #4 above). The flow is fragile.

**9. Hardcoded `time.sleep(10)` (zappa/core.py:~1524)**

A hardcoded 10-second sleep with no explanation. This should either be documented with a reason, or replaced with a polling/waiter approach consistent with the rest of the codebase.

**10. `max()` on potentially empty sequence (zappa/core.py:~1518)**

`max(int(v) for v in versions_in_lambda if v.isdigit())` will raise `ValueError` if there are no numeric versions. Add a guard:
```python
numeric_versions = [int(v) for v in versions_in_lambda if v.isdigit()]
if numeric_versions:
    latest_version = max(numeric_versions)
```

**11. Capacity provider name extraction is fragile (zappa/core.py:~1500-1510)**

The ARN parsing logic (`split("capacity-provider/")` then `rsplit("/")`) is hard to follow and may not handle all formats. Consider a cleaner approach by splitting the ARN by `:` and then parsing the resource component, or using a regex.

**12. Standalone expression `response` does nothing (zappa/core.py:~2802)**

```python
response = self.cloudfront_client.update_distribution(...)
response  # <-- this line does nothing, leftover debug?
```

Remove it.

**13. f-string nested quotes (zappa/core.py:~1516)**

```python
function_arn=f"{response["FunctionArn"]}:{latest_version}"
```
This is a syntax error in Python <3.12. Use single quotes for the dict key inside the f-string:
```python
function_arn=f"{response['FunctionArn']}:{latest_version}"
```

---

### Minor / Style

**14. `log_type="Tail" if not self.capacity_provider_config else "None"` (zappa/cli.py:~1579)**

Why does `log_type` change to `"None"` when using capacity providers? This should have a comment explaining the reasoning — capacity provider log behavior differs from standard Lambda?

**15. `"$LATEST.PUBLISHED"` version identifier (zappa/core.py:~1382)**

This doesn't appear in standard AWS Lambda documentation. If it's capacity-provider-specific, add a comment explaining its purpose so future maintainers understand why it's handled.

**16. Commit history is very messy**

The branch has 80+ commits including many merges from master, unrelated changes (Python 3.11 support, werkzeug fixes, version bumps, etc.), and commits from other contributors' PRs that were merged into the fork. This makes the actual capacity provider changes hard to review. Consider squashing into a clean set of commits that only contain the capacity provider feature.

**17. `function_url_domains` docs added to README but that setting is from a different feature**

The README diff adds `function_url_domains` documentation — this appears to be from the function URL custom domains feature, not the capacity provider feature. If it's already in master, the diff is just noise. If not, it should be a separate PR.

---

### Summary

The core capacity provider integration (config loading, `create_lambda_function`, `update_lambda_configuration`, `wait_for_capacity_provider_response`) is well-implemented with appropriate VPC validation and concurrency guards. The test coverage for these paths is good.

The main blocker is the `_clear_policy()` regression (#1), which will break existing deployments. The missing `return` statements (#2) and return type mismatches (#3, #4) will cause runtime errors. The `Architectures` in `update_function_code` (#6) will cause API errors on every update.

Recommend fixing issues #1-7 before merge, and squashing the commit history.

monkut (commented):
test

PR #1424, opened 2026-02-13
🐛 Fix API Gateway v2 stage handling for custom domains

### LINKED ISSUE #1409: Double stage redirect when using custom domain with API Gateway V2
When using API Gateway V2 with a custom domain (without Route53), Zappa causes a double stage redirect. The custom domain is already mapped to the correct stage in API Gateway's API Mappings, but Zappa still tries to redirect to the stage name, resulting in a duplicated stage path.

For example:
- Custom domain: `api.example.com`
- Stage: `dev`
- API Mapping: `api.example.com` → `dev` stage (path: empty)

When accessing `https://api-test.maquinaweb.com.br/admin/`, it redirects to `https://api-test.maquinaweb.com.br/dev/admin/`, which returns a "not found" error because the stage is duplicated in the path.

Image

This resulting to https://api-test.maquinaweb.com.br/dev/dev/admin/

## Steps to Reproduce

1. Deploy the application with API Gateway V2: `zappa deploy dev`
2. Create a custom domain in API Gateway (without using Route53/Certificate Manager integration from Zappa)
3. Configure API Mapping with empty path pointing to your API and stage
4. Access any endpoint through the custom domain
5. Observe the redirect includes the stage name, causing a double stage path

This my configuration AWS Custom Domain Names with API mappings
Image

## Configuration

zappa_settings.json:

```json
{
    "dev": {
        "aws_region": "sa-east-1",
        "django_settings": "core.settings",
        "exclude": [
            "boto3",
            "botocore",
            "s3transfer",
            "concurrent",
            ".env",
            ".venv",
            ".ruff_cache",
            ".vscode",
            "media"
        ],
        "profile_name": "myprofile",
        "project_name": "myproject",
        "architecture": "arm64",
        "runtime": "python3.11",
        "s3_bucket": "myproject-dev",
        "environment_variables": {
            "PRODUCTION": "True"
        },
        "memory_size": 1024,
        "apigateway_version": "v2",
        "base_path": ""
    }
}
```

Repository to replication: https://github.com/maquinaweb/test-zappa

## Environment

- Zappa version: 0.61.3
- Python version: 3.11
- AWS Region: sa-east-1
- Architecture: arm64
- API Gateway version: v2
- Custom domain configured manually (not through Zappa)

## What I've Tried

1. Setting `"base_path": "/"` in zappa_settings.json - But changed not redirects but all routes is 404
2. Ensuring API Mapping path is empty - Still redirects with stage name
3. Disabling the default API Gateway endpoint - Still redirects

## Additional Context

This issue started occurring after migrating from API Gateway V1 (REST API) to API Gateway V2 (HTTP API). With V1, the custom domain worked correctly without the double stage redirect issue.

Is there a way to disable the automatic stage redirect behavior when using a custom domain with API Gateway V2?

SMCodesP:
## Description

Fix double stage redirect when using API Gateway V2 with a custom domain.

**Problem:** API Gateway V2 always includes the stage name in `rawPath`, even when the request arrives via a custom domain (e.g., `api.example.com`). The previous code always set `SCRIPT_NAME=/{stage}` whenever `requestContext.stage` existed, causing Django/Flask to generate redirect URLs with the stage prefix (e.g., `/dev/admin/`). Since the custom domain's API Mapping already resolves to the correct stage, this resulted in a doubled path like `/dev/dev/admin/`.

**Root cause:** The v1 handler already handled this correctly by checking if the host contains `amazonaws.com` before adding the stage to `SCRIPT_NAME`. The v2 handler was missing this distinction.

**Fix:** Two changes in the v2 event handler (`handler.py`):

1. **Host-based detection:** Only set `SCRIPT_NAME=/{stage}` when the host is `amazonaws.com` (direct API Gateway access). For custom domains, `SCRIPT_NAME` remains empty — consistent with how the v1 handler works.

2. **Manual PATH_INFO stripping:** For custom domain requests, manually strip the stage prefix from `PATH_INFO` after `create_wsgi_request` returns. This is necessary because `create_wsgi_request` only strips the stage when `script_name` is set, but API Gateway V2 always includes the stage in `rawPath` regardless of access method.

| Scenario | `SCRIPT_NAME` | Stage stripped from `PATH_INFO` |
|---|---|---|
| Direct API Gateway (amazonaws.com) | `/{stage}` | By `create_wsgi_request` |
| Custom domain | `""` (empty) | Manually in handler |

**Tests:** Added two new test cases covering the custom domain scenario with realistic API Gateway V2 event payloads. All 26 handler tests pass.

## GitHub Issues

Fixes https://github.com/zappa/Zappa/issues/1409

### DIFF
--- a/tests/test_handler.py
+++ b/tests/test_handler.py
@@ -731,7 +731,7 @@ class TestZappa(unittest.TestCase):
             "https://api.example.com/return/request/url",
         )
 
-    def test_v2_custom_domain_redirect_URL_does_not_duplicate_stage(self):
+    def test_v2_custom_domain_redirect_url_does_not_duplicate_stage(self):
         """
         Verifies WSGI environ values when using a custom domain with API Gateway V2.
 
--- a/tests/test_handler.py
+++ b/tests/test_handler.py
@@ -689,3 +689,82 @@ class TestZappa(unittest.TestCase):
         self.assertIn("SCRIPT_NAME='/dev'", response["body"])
         self.assertIn("PATH_INFO='/debug/wsgi/environ'", response["body"])  # FIXED: stage stripped
         self.assertIn("/dev/debug/wsgi/environ", response["body"])  # Correct single prefix!
+
+    def test_v2_custom_domain_no_double_stage_redirect(self):
+        """
+        Reproduces double stage redirect when using custom domain with API Gateway V2.
+
+        API Gateway V2 always includes the stage in rawPath, even when the request
+        arrives via a custom domain. For direct access (amazonaws.com), SCRIPT_NAME
+        should be set to the stage. For custom domains, SCRIPT_NAME should be empty
+        and the stage should be stripped from PATH_INFO so the framework generates
+        clean URLs without the stage prefix.
+        """
+        lh = LambdaHandler("tests.test_wsgi_script_name_settings")
+
+        # Simulate API Gateway V2 event arriving via custom domain
+        # rawPath INCLUDES the stage prefix (API Gateway V2 always adds it)
+        event = {
+            "version": "2.0",
+            "routeKey": "$default",
+            "rawPath": "/dev/return/request/url",  # Stage prefix IS in rawPath
+            "rawQueryString": "",
+            "headers": {
+                "host": "api.example.com",  # Custom domain, NOT amazonaws.com
+            },
+            "requestContext": {
+                "http": {
+                    "method": "GET",
+                    "path": "/dev/return/request/url",
+                },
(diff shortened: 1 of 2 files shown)


monkut:
Thanks for the diagnosis and fix here — the same root cause (#1409) has been addressed in #1460, which is now merged to `master`. #1460 detects the custom-domain case from `rawPath` (absent stage) and matches the stage on a path-segment boundary, which avoids depending on the `amazonaws.com` host string and the manual `PATH_INFO` stripping. Closing as superseded by #1460.

PR #1431, opened 2026-03-10
Fix `zappa init` including app_function for Django projects

### LINKED ISSUE #1404: `zappa init` error for Django projects
zappa 0.61.2 doesn't generate proper settings for Django which results in
`Error: Warning! Status check on the deployed lambda failed. A GET request to '/' yielded a 502 response code.` during `zappa deploy`

commit https://github.com/zappa/Zappa/commit/a12c4c2bdedc92890368a75e3b7268793f4b8852 introduced `app_function` in base configuration : https://github.com/zappa/Zappa/blob/01962ba6413159986d1ea4b62fc400aa0517a52c/zappa/cli.py#L2189-L2191

which should not be present for a Django project : https://github.com/zappa/Zappa/blob/01962ba6413159986d1ea4b62fc400aa0517a52c/zappa/cli.py#L2034-L2037 when running `zappa init` to create a Django project

monkut:
## Summary
- Fixes #1404 — `zappa init` generated settings with both `app_function` and `django_settings` for Django projects, causing 502 errors on deploy
- `_generate_settings_dict()` always includes `app_function` in the base config; the `init()` flow now removes it when `has_django` is true
- Added regression test that mocks a Django project init and asserts `app_function` is absent from the output

## Test plan
- [x] New test `test_zappa_init_django_excludes_app_function` confirms the fix
- [x] Existing `test_zappa_init` still passes
- [x] All pre-commit hooks pass (isort, black, flake8, mypy)

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -2106,6 +2106,8 @@ class ZappaCLI:
 
         if has_django:
             zappa_settings[env]["django_settings"] = django_settings
+            # django_settings and app_function are mutually exclusive;
+            # remove the default app_function added by _generate_settings_dict()
             zappa_settings[env].pop("app_function", None)
         else:
             zappa_settings[env]["app_function"] = app_function


zappa/cli.py
```diff
@@ -2106,6 +2106,7 @@ def init(self, settings_file: str = "zappa_settings.json"):
 
         if has_django:
             zappa_settings[env]["django_settings"] = django_settings
+            zappa_settings[env].pop("app_function", None)
```
monkut, line 2111: Add comment on to why it's ok to `pop` this.

monkut: Added inline comment explaining the pop: django_settings and app_function are mutually exclusive, removing the default added by `_generate_settings_dict()`.

PR #1435, opened 2026-03-11
:bug: Fix exclude_glob to support ** recursive patterns

### LINKED ISSUE #1269: exclude_glob doesn't respect ** operator
## Context

One of the more useful tools in `glob` is being able to use the `**` operator to match patterns at any point in the tree; however, that requires the `recursive` flag to be turned on in the `glob.glob` function: https://docs.python.org/3.9/library/glob.html#glob.glob

It appears that that has not been used:
https://github.com/zappa/Zappa/blob/f2f03ba8b58c9e2bdbbff7b5e8c905e3a8c08021/zappa/core.py#L638
and 
https://github.com/zappa/Zappa/blob/f2f03ba8b58c9e2bdbbff7b5e8c905e3a8c08021/zappa/core.py#L770

However, the `**` operator is used in the example in the docs: 
https://github.com/zappa/Zappa/blame/f2f03ba8b58c9e2bdbbff7b5e8c905e3a8c08021/README.md#L951

## Expected Behavior
An `exclude_glob` entry with a `**`, e.g. `"**/__pycache__"` should recurse through the directory tree to identify and eliminate matching paths

## Actual Behavior
`**` operators are ignored by `glob`

## Possible Fix
Add the `recursive=True` argument to each of the places where `exclude_glob` paths are used

## Steps to Reproduce
1.  Identify a directory common to multiple paths in the project 
2. Add a matching glob path to `exclude_glob`
3. Run `zappa package`
4. Observe that the paths have not been removed from the packaged project

monkut:
## Summary

Fixes #1269

- `exclude_glob` patterns containing `**` (e.g., `tests/**/*`, `**/*.pyc`) silently matched nothing because the code constructed a literal path from the pattern and then globbed `*` inside it
- Changed to use `temp_project_path.glob(pattern)` which correctly interprets `**` as recursive matching
- Absolute paths in `exclude_glob` are now skipped with a warning (they never matched before either)
- Applied the fix in both locations in `create_lambda_zip()` (project copy and dependency cleanup phases)

## Test plan

- [x] Added `test_exclude_glob_recursive_pattern` test verifying `**/*.pyc` pattern excludes nested `.pyc` files
- [x] Existing `test_get_exclude_glob__file_not_deleted` still passes (absolute path edge case)
- [x] Full test suite passes (263 tests)

### DIFF
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -15,6 +15,7 @@ from zappa.utilities import (
     ApacheNCSAFormatter,
     EventSourceMappingMixin,
     InvalidAwsLambdaName,
+    S3EventSource,
     conflicts_with_a_neighbouring_module,
     contains_python_files_or_subdirs,
     detect_django_settings,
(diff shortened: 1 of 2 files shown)


zappa/core.py
```diff
@@ -570,8 +570,10 @@ def splitpath(path):
             else:
                 copytree(cwd, temp_project_path, metadata=False, symlinks=False)
             for glob_path in exclude_glob:
-                exclude_glob_path = temp_project_path / glob_path
-                for path in exclude_glob_path.glob("*"):
+                if Path(glob_path).is_absolute():
```
monkut, line 575: Why are relative paths required here?

zappa/core.py
```diff
@@ -672,8 +674,10 @@ def splitpath(path):
 
         # Cleanup
         for glob_path in exclude_glob:
-            exclude_glob_path = temp_project_path / glob_path
-            for path in exclude_glob_path.glob("*"):
+            if Path(glob_path).is_absolute():
+                logger.warning(f"exclude_glob: skipping absolute pattern '{glob_path}' (use relative patterns)")
```
monkut, line 682: Why are relative paths required here?

zappa/core.py
monkut: Patterns must be relative because `temp_project_path.glob(pattern)` resolves them against the temp directory where the project is copied. An absolute path (e.g. `/home/user/foo`) would never match files inside the temp copy. Added an inline comment explaining this.

zappa/core.py
monkut: Same reason as above — `temp_project_path.glob(pattern)` resolves patterns relative to the temp directory, so absolute paths can never match. Added an inline comment here as well.

PR #1439, opened 2026-03-11
:bug: Fix resource policy limit with many scheduled events

### LINKED ISSUE #1415: API Gateway Resource Policy limits with too many scheduled events
Context
On a recent deployment, I hit this error:
```
An error occurred (PolicyLengthExceededException) when calling the AddPermission operation: The final policy size (20698) is bigger than the limit (20480).
```
When looking at the policy, I can see that Zappa is creating one policy entry for each task, e.g:

```
{
  "StringEquals": {
    "AWS:SourceAccount": "$AWS_ACCOUNT_ID"
  },
  "ArnLike": {
    "AWS:SourceArn": "arn:aws:events:$AWS_REGION:$AWS_ACCOUNT_ID:rule/$AWS_LAMBDA_FUNCTION_NAME.$APP_TASK_NAME"
  }
}
```

I know this is an AWS limit and not strictly a Zappa issue, but I wonder if there can be a more efficient way to handle these permissions so that larger functions can support more events.

Possible Fix
Either one of these:

a) (preferred, better UX for users) a policy generator that uses a wildcard in the SourceArn so that all events can be supported by a single clause in the policy

b) (easier implementation for Zappa) a way to disable automatic API Gateway Resource Policy management in Zappa so that users can manage the policy manually

monkut:
## Summary

Fixes #1415

- Each scheduled event called `create_event_permission()` with a specific `SourceArn`, adding a new policy statement per rule
- With many events, this caused the Lambda resource policy to exceed AWS's 20KB limit
- Replaced per-rule permissions with a single wildcard `arn:aws:events:{region}:{account}:rule/*` permission added once before scheduling events
- `_clear_policy()` already removes old per-rule statements on each `zappa update`, so the migration from per-rule to wildcard is seamless
- The wildcard permission still uses `SourceAccount` to restrict invocations to the same AWS account

## Test plan

- [x] Full test suite passes (271 tests, 2 pre-existing websocket flakes)
- [ ] Manual testing: deploy with 50+ scheduled events to verify policy stays under 20KB


zappa/core.py
```diff
@@ -3307,6 +3307,28 @@ def create_event_permission(self, lambda_name, principal, source_arn):
...
+            self.lambda_client.add_permission(
+                FunctionName=lambda_name,
+                StatementId="ZappaEventBridgeWildcard",
+                Action="lambda:InvokeFunction",
+                Principal="events.amazonaws.com",
+                SourceArn=f"arn:aws:events:{region}:{account_id}:rule/*",
+                SourceAccount=account_id,
+            )
+        except self.lambda_client.exceptions.ResourceConflictException:
+            # Permission already exists
```
monkut, line 3329: Add logging that: 'permission already exists'

zappa/core.py
```diff
@@ -3332,6 +3354,13 @@ def schedule_events(self, lambda_arn, lambda_name, events, default=True):
             events=events,
             excluded_source_services=pull_services,
         )
+
+        # Add a single wildcard permission for all EventBridge rules instead of
+        # one per rule, to stay within the 20KB resource policy limit (#1415).
+        has_scheduled_events = any(e.get("expression") or e.get("expressions") for e in events)
+        if has_scheduled_events:
```
monkut, line 3361: if has_scheduled_events is false, is it necessary to then loop through `events`.

PR #1440, opened 2026-03-11
:bug: Fix double stage redirect with V2 custom domain mapping

### LINKED ISSUE #1409: Double stage redirect when using custom domain with API Gateway V2

monkut:
## Summary

Fixes #1409

- When a custom domain maps to an API Gateway v2 stage, API Gateway strips the stage prefix from `rawPath` before invoking Lambda
- But the handler always set `SCRIPT_NAME` to `/{stage}`, causing Django/Flask to construct redirect URLs with a duplicated stage (e.g., `/dev/dev/admin/`)
- Now detects whether `rawPath` actually starts with `/{stage}` — if not (custom domain case), `SCRIPT_NAME` is left empty
- Direct API Gateway v2 access (where `rawPath` includes the stage) continues to work as before

## Test plan

- [x] Added `test_wsgi_v2_custom_domain_no_double_stage` — verifies custom domain event produces correct URL without stage prefix
- [x] Existing `test_wsgi_script_name_on_v2_formatted_event_with_stage_in_path` still passes (direct API Gateway access)
- [x] All 25 handler tests pass
- [x] Full test suite passes (272 tests)

### DIFF
--- a/tests/test_handler.py
+++ b/tests/test_handler.py
@@ -689,3 +689,40 @@ class TestZappa(unittest.TestCase):
         self.assertIn("SCRIPT_NAME='/dev'", response["body"])
         self.assertIn("PATH_INFO='/debug/wsgi/environ'", response["body"])  # FIXED: stage stripped
         self.assertIn("/dev/debug/wsgi/environ", response["body"])  # Correct single prefix!
+
+    def test_wsgi_v2_custom_domain_no_double_stage(self):
+        """
+        Test that API Gateway v2 with a custom domain mapping doesn't double the
+        stage in URLs (#1409). When a custom domain maps to a stage, API Gateway
+        strips the stage from rawPath, so SCRIPT_NAME should be empty.
+        """
+        lh = LambdaHandler("tests.test_wsgi_script_name_settings")
+
+        event = {
+            "version": "2.0",
+            "routeKey": "$default",
+            "rawPath": "/return/request/url",  # Custom domain: stage already stripped
+            "rawQueryString": "",
+            "headers": {
+                "host": "api.example.com",
+            },
+            "requestContext": {
+                "http": {
+                    "method": "GET",
+                    "path": "/return/request/url",
+                },
+                "stage": "dev",  # Stage is still present in requestContext
+                "domainName": "api.example.com",
+            },
+            "isBase64Encoded": False,
+            "body": "",
+        }
+        response = lh.handler(event, None)
+
+        self.assertEqual(response["statusCode"], 200)
+        # With custom domain, stage is NOT in rawPath, so SCRIPT_NAME should be empty
+        # and the URL should NOT contain /dev prefix
+        self.assertEqual(
+            response["body"],
+            "https://api.example.com/return/request/url",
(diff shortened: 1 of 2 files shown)


monkut:
Superseded by #1460, which is now merged to `master`. #1460 keeps this fix and hardens the stage detection to match on a path-segment boundary (exact `/{stage}` or `/{stage}/` prefix) instead of a bare string prefix, so a custom-domain route that merely shares the stage's leading substring (e.g. stage `prod` + `/products`) is no longer mis-stripped. Closing in favor of #1460.

PR #1467, opened 2026-07-20
Fix merge_headers() joining multiple Cookie headers with ", "

### LINKED ISSUE #1466: merge_headers() joins multiple Cookie headers with ", " instead of "; ", breaking Django sessions over HTTP/2
## Bug

`merge_headers()` joins every multi-value header with `", "`:

https://github.com/zappa/Zappa/blob/master/zappa/utilities.py#L965-L966

That's fine for most headers, but cookies are separated by `"; "`, not commas.

HTTP/2 clients (Chrome, for one) may send the cookie header as multiple field lines, 
which API Gateway payload v1 delivers as a list in `multiValueHeaders["cookie"]`.
Zappa then reassembles them as `csrftoken=A, sessionid=B` — Django only splits cookies on `";"`, so `sessionid` is silently dropped and the user becomes anonymous.
The result is a login redirect loop that's hard to debug, since browser DevTools still shows one well-formed `Cookie:` line.

## Reproduction

```python
from zappa.utilities import merge_headers

event = {
    "headers": {"cookie": "csrftoken=AAAA"},
    "multiValueHeaders": {"cookie": ["csrftoken=AAAA", "sessionid=BBBB"]},
}
merge_headers(event)
# actual:   {'cookie': 'csrftoken=AAAA, sessionid=BBBB'}
# expected: {'cookie': 'csrftoken=AAAA; sessionid=BBBB'}
```

Environment: Zappa 0.62.1 (still present on `master`), API Gateway REST API (payload v1), Django 5.x, Chrome 150.

## Suggested fix

```python
for h in multi_headers.keys():
    separator = "; " if h.lower() == "cookie" else ", "
    multi_headers[h] = separator.join(multi_headers[h])
```

The payload v2 path already joins `event["cookies"]` with `"; "` (`zappa/wsgi.py`), so this just brings v1 in line with v2.

Happy to open a PR with the fix and a regression test.

kimgunnn:
## Description

Fixes #1466

`merge_headers()` joins all multi-value headers with `", "`, 
including `Cookie` — but cookies are separated by `"; "`, not commas.
HTTP/2 clients such as Chrome may split the cookie header into multiple field lines, 
which API Gateway payload v1 delivers via `multiValueHeaders`.
Zappa then reassembles them as `csrftoken=A, sessionid=B`, 
which WSGI frameworks cannot parse — `sessionid` is silently dropped and session auth breaks (e.g. Django login loops).

The payload v2 path already joins `event["cookies"]` with `"; "`
(`zappa/wsgi.py`), so this change brings v1 in line with v2.

## GitHub Issues

#1466

### DIFF
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -312,12 +312,16 @@ USE_TZ = True
     def test_merge_headers_joins_cookie_with_semicolon(self):
         # HTTP/2 clients may send the cookie header as multiple field lines;
         # they must be re-joined with "; ", not ", ", for WSGI frameworks to parse them.
-        event = {
-            "headers": {"Cookie": "csrftoken=AAAA"},
-            "multiValueHeaders": {"Cookie": ["csrftoken=AAAA", "sessionid=BBBB"]},
-        }
-        merged = merge_headers(event)
-        self.assertEqual(merged["Cookie"], "csrftoken=AAAA; sessionid=BBBB")
+        for header_name in ("Cookie", "cookie"):
+            with self.subTest(header_name=header_name):
+                event = {
+                    "headers": {header_name: "csrftoken=AAAA"},
+                    "multiValueHeaders": {
+                        header_name: ["csrftoken=AAAA", "sessionid=BBBB"]
+                    },
+                }
+                merged = merge_headers(event)
+                self.assertEqual(merged[header_name], "csrftoken=AAAA; sessionid=BBBB")
 
     def test_merge_headers_joins_other_headers_with_comma(self):
         event = {
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -23,6 +23,7 @@ from zappa.utilities import (
     get_venv_from_python_version,
     human_size,
     is_valid_bucket_name,
+    merge_headers,
     parse_s3_url,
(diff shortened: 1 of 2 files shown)


monkut (changes requested):
The fix itself is correct and verified.
Requesting a few small changes to align with project conventions before merge:

1. **Add a lowercase `cookie` test case.** The new test uses `"Cookie"`, but the real-world HTTP/2 shape (and the reproduction in #1466) is lowercase `"cookie"` — which is exactly what the `h.lower()` branch exists to handle, and it's currently untested. A second assertion or a parametrized case with `{"multiValueHeaders": {"cookie": [...]}}` would cover it.

Optional, non-blocking: the existing `merge_headers` tests live in `tests/test_handler.py` (`test_merge_headers_no_multi_value` etc.). Since the function lives in `zappa/utilities.py`, the new tests' placement in `tests/test_utilities.py` is reasonable — but consider moving the old ones alongside them in a follow-up so coverage of the function isn't split across two files.

kimgunnn:
Added the lowercase case via subTest so both spellings are covered. Re: consolidating the existing merge_headers tests from test_handler.py, happy to do that in a follow-up PR.

Issue #1463, opened 2026-07-08
Snapstart requires invoking an alias or version, not $LATEST


dfuentes77:
Referring to the documentation:
https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html#snapstart-runtimes

`
You can use SnapStart only on published function versions and aliases that point to versions. You can't use SnapStart on a function's unpublished version ($LATEST).
`

But Zappa defaults to invoking $LATEST in the API-gateway.  Am I missing a setting where it updates this appropriately?

PR #1468, opened 2026-07-22
full snapstart and provisioned concurrency support

dfuentes77:
## Description

Hoping to support Snapstart and Provisioned Concurrency while using aliases, since they are required.  

This also allows the deployments to be more blue/green by having ALB and api-gateway point to aliases and only migrating an alias when newly deployed lambda is "ready"
  
New tests added.  All tests pass locally.

## GitHub Issues

https://github.com/zappa/Zappa/issues/1463

### DIFF
--- a/README.md
+++ b/README.md
@@ -1180,6 +1180,7 @@ to change Zappa's behavior. Use these at your own risk!
         // If you are adding a Function URL manually (e.g. outside Zappa) and seeing 403s with NONE
         // auth, this two-statement shape is the missing piece.
         "apigateway_version": "v1", // optional, API Gateway version to use. Can be "v1" or "v2". Default "v1".
+        "apigateway_lambda_qualifier": "live", // optional, Lambda version or alias for API Gateway integrations. Default null uses the unqualified function ARN, except when `snap_start` or `provisioned_concurrency` is enabled, in which case it defaults to Zappa's managed alias for that feature (see `snap_start`/`provisioned_concurrency` below). Set this only to point at your own alias instead.
         "architecture": "x86_64", // optional, Set Lambda Architecture, defaults to x86_64. For Graviton 2 use: arm64
         "async_source": "sns", // Source of async tasks. Defaults to "lambda"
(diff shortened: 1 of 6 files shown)


monkut:
Took a look at this along with #1463 — the issue is real. On master we publish versions for SnapStart, but API Gateway still points at the unqualified ARN, so everything invokes `$LATEST` and the snapshots never actually get used. The alias approach here is the right fix.

I checked out the branch and ran the test suite locally — all green (the few failures I hit were just my machine running out of disk, they fail on master too). With none of the new settings configured the generated templates and Lambda calls are unchanged, so existing deployments aren't affected.

CI is only failing on `black --check` — running `make black` should sort it out. The actual test jobs never ran.

A few things I noticed:

- One that I think is worth addressing in this PR: on `zappa update`, the provisioned-concurrency version gets published (and the alias moved to it) *before* the config update lands on `$LATEST`. Since a published version freezes code + config together, env var / memory / timeout changes won't reach alias-routed traffic until the *next* update. That's likely to confuse people — you change an env var, run `zappa update`, the deploy succeeds, but live traffic still sees the old value with no indication why. Proposed fix: move the provisioned-concurrency handling (publish, `put_provisioned_concurrency_config`, wait for READY, migrate alias, clean up the old version's config) out of `update_lambda_function` and into `update_lambda_configuration`, after the config change is applied — i.e. the same ordering the SnapStart path already uses. Then the published version captures both the new code and the new config in one deploy.
And two smaller ones, not blockers:

- With a very low `num_retained_versions` (like 1), version pruning can try to delete the version the `snapstart` alias still points to and blow up with `ResourceConflictException`. The default of 5 is fine, but maybe worth a guard or a doc note.
- `zappa rollback` republishes old code but doesn't move the aliases, so traffic won't actually roll back for SnapStart/provisioned-concurrency users. The ALB alias has the same gap today, so fine as a follow-up.

dfuentes77:
now that [AWS supports SnapStart for Lambda containers](https://aws.amazon.com/about-aws/whats-new/2026/07/aws-lambda-snapstart-container/), I wanted to ensure support for that in this PR

PR #1473, opened 2026-08-28
:bug: Use rawQueryString for API Gateway v2 to preserve repeated query params

### LINKED ISSUE #1472: API Gateway v2 (HTTP API) collapses repeated query parameters into one comma-joined value
## Context

First off — thanks for Zappa, and for the recent API Gateway v2 work. Moving our project to `"apigateway_version": "v2"` was almost entirely painless, so I hope this report is useful rather than a nuisance.

We hit one snag: with `v2`, a request containing repeated query parameters (`?id=18&id=19&id=20`) arrives at the application as a **single** parameter with the value `"18,19,20"`. The same request under `v1` works correctly.

This is because payload format 2.0 doesn't include the `multiValueQueryStringParameters` field that format 1.0 and ALB provide. Instead
it gives you two things:

- `queryStringParameters` — repeats flattened into a comma-joined value, i.e.  `{"id": "18,19,20"}`
- `rawQueryString` — the query string verbatim and percent-encoded, i.e.  `"id=18&id=19&id=20"`

[`process_lambda_payload_v2()`](https://github.com/zappa/Zappa/blob/master/zappa/wsgi.py#L167-L168) currently reads the first one:

```python
query = event_info.get("queryStringParameters", {})
query_string = urlencode(query) if query else ""
```

`urlencode({"id": "18,19,20"})` produces `id=18%2C19%2C20`, which frameworks parse back as a single value. The v1 path handles this correctly just above, via `urlencode(query, doseq=True)` over `multiValueQueryStringParameters`.

A couple of related notes that might be worth folding into the same fix:

- ASGI is affected too — [`zappa/asgi.py`](https://github.com/zappa/Zappa/blob/master/zappa/asgi.py#L25) calls the same function, so a single change covers both.
- Lambda Function URLs are affected as well, since they also use payload format 2.0 regardless of the `apigateway_version` setting.
- Even for a parameter that isn't repeated, `queryStringParameters` can't tell `?tags=a,b` (one value with a comma) from `?tags=a&tags=b` whereas `rawQueryString`can.

API Gateway v1 and ALB are unaffected.

## Expected Behavior

Repeated parameters reach the application intact, identically under `v1` and `v2`:

```python
request.GET.getlist("id")    # ['18', '19', '20']
```

## Actual Behavior

Under `v2` the application receives one value:

```python
request.GET.getlist("id")    # ['18,19,20']
```

In our case this surfaced as a 500 from a Wagtail multiple snippet chooser, which sends selected IDs as repeated `id` parameters and passes them to a Django `__in` filter:

```
ValueError at /admin/snippets/choose/tabs/licenceclass/chosen-multiple/
Field 'id' expected a number but got '18,19,20,26,10,21,28,23,11,22'.
```

The same would apply to `request.args.getlist()` in Flask, `django-filter`, DRF filter backends, and anything else built on repeated parameters.

## Possible Fix

Build `QUERY_STRING` from `rawQueryString` when that key is present, passing it through as-is — API Gateway already delivers it percent-encoded, so re-encoding would double-escape it and unquoting would break a literal `&` or `=` inside a value. Falling back to `queryStringParameters` when the key is absent keeps non-API-Gateway invokers working.

I have a PR ready with this change plus regression tests, and I'm happy to adjust the approach if you'd prefer to solve it differently.

One small thing that may explain how this slipped through: every v2 event fixture in the test suite currently uses `"rawQueryString": ""`, and the only v2 assertion on `QUERY_STRING` checks that it equals `""` — so it passes trivially.
There's no existing test that sends a v2 event with a non-empty query string.

## Steps to Reproduce

No AWS account needed — this reproduces against the installed package, using the event shape API Gateway v2 actually sends.

1. Save as `repro.py`:

```python
from urllib.parse import parse_qs

from zappa.wsgi import create_wsgi_request

# What API Gateway v2 sends for GET /thing?id=18&id=19&id=20
v2_event = {
    "version": "2.0",
    "routeKey": "ANY /{proxy+}",
    "rawPath": "/thing",
    "rawQueryString": "id=18&id=19&id=20",
    "queryStringParameters": {"id": "18,19,20"},  # <- AWS flattens repeats here
    "headers": {"host": "example.com"},
    "requestContext": {"http": {"method": "GET", "path": "/thing"}, "stage": "dev"},
    "isBase64Encoded": False,
}

# The equivalent v1 event
v1_event = {
    "httpMethod": "GET",
    "path": "/thing",
    "headers": {"Host": "example.com"},
    "multiValueQueryStringParameters": {"id": ["18", "19", "20"]},
    "requestContext": {},
    "body": None,
}

for name, event in (("v2", v2_event), ("v1", v1_event)):
    qs = create_wsgi_request(event)["QUERY_STRING"]
    print(f"{name}: QUERY_STRING={qs!r}  parse_qs={parse_qs(qs)}")
```

2. Run `python repro.py`

3. Output:

```
v2: QUERY_STRING='id=18%2C19%2C20'   parse_qs={'id': ['18,19,20']}
v1: QUERY_STRING='id=18&id=19&id=20' parse_qs={'id': ['18', '19', '20']}
```

The `v1` line is what both should produce.

## Your Environment

- Zappa version used: latest (main branch)
- Operating System and Python version: Python 3.12 Docker based Lambda
- Your `zappa_settings.json`:

```json
{
    "production": {
        "apigateway_version": "v2"
    }
}
```

Thanks again for maintaining this — happy to test any fix against our deployment.

alexalligator:
## Description

API Gateway payload format 2.0 has no multiValueQueryStringParameters and flattens repeated parameters into a comma-joined value in queryStringParameters (?id=1&id=2 => {"id": "1,2"}). process_lambda_payload_v2 read that lossy field, so applications saw a single value "1,2" instead of two, breaking anything relying on repeated parameters (Wagtail choosers, request.GET.getlist, django-filter, DRF).

Build QUERY_STRING from rawQueryString instead, which API Gateway delivers verbatim and already percent-encoded, so it is passed through without urlencode (double-escaping) or unquote (which would destroy a literal & or = inside a value). This also preserves a literal comma in a single value, which queryStringParameters cannot express.

The presence check is "rawQueryString" in event_info rather than truthiness, so an empty query string does not fall through to the legacy path. Invokers other than API Gateway may send a v2-shaped event with no rawQueryString; that fallback now uses doseq=True to match the v1 multi-value branch.

zappa/asgi.py calls the same function, so ASGI apps are fixed too, as are Lambda Function URLs, which use payload format 2.0 regardless of the apigateway_version setting. v1 (REST API) and ALB are unchanged.

This was invisible to the suite because every v2 fixture used "rawQueryString": "" and the only v2 QUERY_STRING assertion checked for "". Adds regression tests at the WSGI, ASGI and handler levels.

## GitHub Issues

Closes #1472

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Fix API Gateway v2 (HTTP API) collapsing repeated query parameters such as `?id=1&id=2` into a single comma-joined value; `QUERY_STRING` is now built from the event's `rawQueryString`, which also fixes ASGI apps and Lambda Function URLs (#1472)
 * Change default of `num_retained_versions` from `null` (keep all) to `5` (#1453)
   - Lambda code storage and SnapStart snapshot-cache cost now have a sane default upper bound.
   - On the first `zappa update` after upgrade, published versions older than the newest 5 are deleted; versions referenced by an alias (e.g. ALB) and `$LATEST` are unaffected, but versions referenced by other aliases can still raise `ResourceConflictException` (see #960).
--- a/tests/test_asgi.py
+++ b/tests/test_asgi.py
@@ -418,6 +418,38 @@ class TestASGI(unittest.TestCase):
         self.assertIn("foo=bar", response["body"])
         self.assertIn("baz=qux", response["body"])
 
+    def test_asgi_v2_with_repeated_query_string(self):
+        """
+        Repeated query parameters must survive the payload format 2.0
+        conversion into the ASGI scope, not be collapsed into "id=18,19,20".
+        https://github.com/zappa/Zappa/issues/1472
+        """
+        lh = LambdaHandler("tests.test_asgi_settings")
+
+        event = {
+            "version": "2.0",
+            "routeKey": "$default",
+            "rawPath": "/return/request/url",
+            "rawQueryString": "id=18&id=19&id=20",
+            # This is the lossy value API Gateway v2 actually sends alongside it.
+            "queryStringParameters": {"id": "18,19,20"},
+            "headers": {
+                "host": "example.com",
+            },
+            "requestContext": {
+                "http": {
+                    "method": "GET",
+                    "path": "/return/request/url",
+                },
+            },
+            "isBase64Encoded": False,
+            "body": "",
+        }
+        response = lh.handler(event, None)
+
+        self.assertEqual(response["statusCode"], 200)
+        self.assertIn("?id=18&id=19&id=20", response["body"])
+
     def test_asgi_404(self):
         """Ensure ASGI app returns 404 for unknown routes."""
         lh = LambdaHandler("tests.test_asgi_settings")
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -18,6 +18,7 @@ from functools import partial
 from io import BytesIO
 from pathlib import Path
 from subprocess import check_output
+from urllib.parse import parse_qs
 
 import botocore
 import botocore.stub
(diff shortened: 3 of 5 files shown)


alexalligator:
Hi, one more thing:

Note that there are existing test failures on `main`. This PR adds 7 new passing tests.

